PR #244, opened 2023-11-28
[Bugfix] Menu bar item not hidden after restart.

NeverAgain11:
add a condition check before adding the status item


NeverAgain11:
#235 修复该问题

tisfeng (approved):

tisfeng:
赞

Issue #287, opened 2023-12-30
【Swift 重构】SwiftUI 补全设置-通用


tisfeng:
<img width="536" alt="image" src="https://github.com/tisfeng/Easydict/assets/25194972/ff8166fb-e479-405f-92d8-4fc36e184917">

<img width="617" alt="image" src="https://github.com/tisfeng/Easydict/assets/25194972/d50650ee-77f9-47c7-b8e6-3104dac07b19">

CanglongCl:
closed with #323 
for shortcuts, track in #290

Issue #221, opened 2023-11-14
使用苹果开源的 OrderedDictionary 替换项目中的 MMOrderedDictionary


tisfeng:
参见 https://github.com/tisfeng/Easydict/pull/218#discussion_r1391908370

tisfeng:
后续如果在项目中看到其他老旧的第三方库，也可以这样开个 issue 进行替换。

Kyle-Ye:
It will be blocked by the current ObjC codebase. We may consider doing so when we finish all the usage's migration to Swift.

Most value types (struct/enum) can't be used in ObjC code.

String, Dictionary and Array is just specially ones which can be bridged to NSString, NSDictionary and  NSArray.

If we replace MMOrderedDictionary with OrderedDictionary, all the ObjC code can not utilize it.

tisfeng:
好吧，那暂时先放着，等我们 Swift 迁移完成后再来替换。

PR #323, opened 2024-01-13
Introduce `Defaults` package into project and implement Setting - General with SwiftUI

CanglongCl:
close https://github.com/tisfeng/Easydict/issues/287


CanglongCl:
I'm not sure how the project store other type in original storage. To use `Defaults` with other type like `EZLanguage`, we need to implement `Defaults.Serializable` for them. What's our target to refactor them? Write a new one in `Swift` or maintain `objc` for a while?

tisfeng:
EZLanguage can be rewritten with Swift, however I am not sure if this would affect the objc code that use EZLanguage, as it is used in many places.

If other objc code will not be affected, then EZLanguage can be rewritten directly. If there is an impact, many places will have to make corresponding changes, which may be a lot of work.

Overall, I am in favor of rewriting EZLanguage with Swift.

CanglongCl:
BTW I think this PR can be just merged. Other related migration can be based on `Defaults` package.

CanglongCl:
image
`GeneralTab` is almost completed except keyboard shortcut section. I've also added the beta feature toggle as mentioned in #321 and #135.

tisfeng:
That’s great, it works well. 

Just two more options are missing: appearance and font size, could you add those please?

image

image

CanglongCl:
image

image

Easydict/NewApp/Configuration/EZConfiguration.swift
```diff
@@ -0,0 +1,55 @@
+//
+//  EZConfiguration.swift
```
tisfeng, line 2: In Swift, maybe we don't have to add `EZ` prefix like objc?

Easydict/NewApp/Model/TTSService.swift
```diff
@@ -0,0 +1,80 @@
...
+//
+//  Created by 戴藏龙 on 2024/1/13.
+//  Copyright © 2024 izual. All rights reserved.
+//
+
+import Defaults
+import Foundation
+
+@available(macOS 13, *)
+enum TTSService: String, CaseIterable, CustomLocalizedStringResourceConvertible {
```
tisfeng, line 13: Can we rename `TTSService` to `TTSServiceType`?  Since it's a service type here, not a service.

Easydict/NewApp/Utility/Extensions/EZLanguageDetectOptimizeExtensions.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  EZLanguageDetectOptimizeExtensions.swift
```
tisfeng, line 2: remove `EZ` prefix.

Easydict/NewApp/Utility/Extensions/EZShowWindowPositionExtensions.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  EZShowWindowPositionExtensions.swift
```
tisfeng, line 2: remove `EZ` prefix.

Easydict/NewApp/Utility/Extensions/EZWindowTypeExtensions.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  EZWindowTypeExtensions.swift
```
tisfeng, line 2: remove `EZ` prefix.

Easydict/NewApp/Utility/Extensions/EZLanguageExtensions.swift
```diff
@@ -0,0 +1,78 @@
...
+
+extension Language: Defaults.Serializable {}
+
+extension Language: CaseIterable {
+    public static let allCases: [Language] = EZLanguageModel.allLanguagesDict().allKeys().map { rawValue in
+        Language(rawValue: rawValue as String)
+    }
+
+    public static let allCasesWithoutAuto: [Language] = allCases.filter { language in
+        language != .auto
```
tisfeng, line 20: User's preferred language should remove `文言文`

image

tisfeng: Can we keep the order of user-preferred languages consistent with before? It is more convenient for users to choose languages in this way.

old setting
image

tisfeng: Currently, we use `MMOrderedDictionary` to manage the order of languages, we may need to replace  `MMOrderedDictionary` with Apple's `OrderedDictionary` latter.

See in https://github.com/tisfeng/Easydict/issues/221

CanglongCl: Any idea to maintain the order?

Easydict.xcodeproj/project.pbxproj
```diff
@@ -3251,6 +3311,14 @@
 				minimumVersion = 5.8.1;
 			};
 		};
+		EA3B81FA2B52555C004C0E8B /* XCRemoteSwiftPackageReference "Defaults" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/sindresorhus/Defaults.git";
+			requirement = {
+				branch = main;
+				kind = branch;
```
tisfeng, line 3319: Why does `Defaults` use branches instead of version numbers?

image

CanglongCl: fixed.

tisfeng (approved):

PR #335, opened 2024-01-17
[Refactor] rewrite EZConfiguration using Swift

NeverAgain11:
#284


CanglongCl:
`@Defaults.Default` property wrapper should be only used in SwiftUI since it introduce a `SwiftUI.ObservedObject` for UI tracking which we don't need here. 

I wrote a property wrapper for you. Have a try. 

```swift 
@propertyWrapper
struct DefaultsWrapper<T: Defaults.Serializable> {
    var wrappedValue: T {
        get {
            Defaults[key]
        } set {
            Defaults[key] = newValue
        }
    }

    /// Initialize a property binding to Defaults value.
    /// - Parameters:
    ///   - key: key in Defaults
    ///   - onUpdated: the function called on value changes
    ///   - callInitial: the function called on initialize the property
    init(_ key: Defaults.Key<T>, onUpdated: @escaping (_ oldValue: T, _ newValue: T) -> Void, onInitial: ((_ value: T) -> Void)?) {
        self.key = key
        if let onInitial {
            onInitial(Defaults[key])
        }
        observer = Defaults.observe(key) { change in
            onUpdated(change.oldValue, change.newValue)
        }
    }

    private let observer: Defaults.Observation

    let key: Defaults.Key<T>
}

@DefaultsWrapper(
    .firstLanguage,
    onUpdated: { _, newValue in
        logSettings(["first_language": newValue])
    },
    onInitial: { logSettings(["first_language": $0]) }
)
var firstLanguage
```

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,595 @@
...
+                end repeat
+                if \(launchAtStartup) then
+                    make login item at end with properties {path:"\(appBundlePath)", hidden:false}
+                end if
+            end tell
+        """
+
+        let exeCommand = EZScriptExecutor()
+        exeCommand.runAppleScript(script) { result, error in
+            if error != nil {
```
CanglongCl, line 584: Compiler says `Comparing non-optional value of type 'EZError' to 'nil' always returns true`

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,595 @@
...
+class Configuration: NSObject {
+    private(set) static var shared = Configuration()
+
+    var appDelegate = NSApp.delegate as? AppDelegate
+
+    var updater: SPUUpdater? {
+        appDelegate?.updaterController.updater
+    }
+
+    @Default(.firstLanguage)
```
CanglongCl, line 21: `@Defaults.Default` property wrapper should be only used in SwiftUI since it introduce a `SwiftUI.ObservedObject` inside for UI tracking which we don't need here.

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    func didSetAppearance() {
+        DarkModeManager.sharedManager().updateDarkMode()
+    }
+}
+
+// MARK: Window Frame
+
+private extension Configuration {
```
CanglongCl, line 480: Try use `Defaults.Bridge` to comfort `CGRect` to `Serializable` and introduce a Defaults.Key for it?

CanglongCl (changes requested):

CanglongCl:
I implement `intelligentQueryTextType` for your reference. 

```swift
extension Defaults.Keys {
    static func intelligentQueryTextType(for serviceType: ServiceType) -> Key<EZQueryTextType> {
        let key = EZConstKey.constkey("IntelligentQueryTextType", serviceType: serviceType)
        return .init(key, default: EZQueryTextType(rawValue: 7))
    }
}

extension EZQueryTextType: Defaults.Serializable {
    public static var bridge: Bridge = Bridge()

    public struct Bridge: Defaults.Bridge {
        public func serialize(_ value: EZQueryTextType?) -> String? {
            guard let value else { return "7" }
            return "\(value.rawValue)"
        }
        
        public func deserialize(_ object: String?) -> EZQueryTextType? {
            guard let object else { return nil }
            return EZQueryTextType(rawValue: UInt(object) ?? 7)
        }
        
        public typealias Value = EZQueryTextType

        public typealias Serializable = String
    }
}
```

Use with 

```swift
Defaults[.intelligentQueryTextType(for: .ali)]
```

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,595 @@
...
+    func windowFrameWithType(_ windowType: EZWindowType) -> CGRect {
```
CanglongCl, line 481: Try use `Defaults.Bridge` to comfort `CGRect` to `Serializable` and introduce a Defaults.Key for it?

CanglongCl (changes requested):

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    @Default(.firstLanguage)
+    var firstLanguage: Language
+
+    @Default(.secondLanguage)
+    var secondLanguage: Language
+
+    @Default(.queryFromLanguage)
+    var from: Language
```
phlpsong, line 28: It would be better to rename `from` and `to` to a more readable name like `firstLanguage` above.

NeverAgain11:
@CanglongCl 
DefaultsWrapper has an error.
image

CanglongCl:
> @CanglongCl DefaultsWrapper has an error. image

Try make `func logSettings(_ parameters: [String: Any])` to `static func logSettings(_ parameters: [String: Any])` and make `func didSetFirstLanguage()` to `static func didSetFirstLanguage()`

NeverAgain11:
Defaults.observe already meet the need for observation.

tisfeng:
@NeverAgain11 Please resolve conflicts.

AkaShark (approved):

NeverAgain11:
> @NeverAgain11 Please resolve conflicts.

Done.

tisfeng:
@NeverAgain11  This code looks fine, but we need to test their functionality in real use.

After rewriting EZConfiguration, you should try to completely replace the previous objc EZConfiguration with the new Swift Configuration and make sure that both the new SwiftUI settings page and the previous settings page work properly.

In addition, I noticed that both `NewConfiguration.swift` and `Configuration.swift` exist, maybe we need to merge them and move the Swift code all under a new group, moving to the `Configuration.swift` file.

tisfeng:
I'm not sure exactly what you are referring to, can you elaborate using code? Is it possible to use `Class` instead of `Struct`? Whatever the problem is, we need to figure out how to fix it.

If some of the Swift properties really don't work in objc, we can start by replacing `EZConfiguration` in Swift/SwiftUI with `Configuration.swift` and make sure that the new settings page functionality works fine. `EZConfiguration` in objc can be left unchanged for now.

tisfeng:
I tried that, and it's probably not really a good idea to completely replace objc `EZConfiguration` with `Configuration.swift`, so let's just go ahead and replace the one in the Swift code.

CanglongCl:
> I'm not sure exactly what you are referring to, can you elaborate using code? Is it possible to use `Class` instead of `Struct`? Whatever the problem is, we need to figure out how to fix it.
> 
> If some of the Swift properties really don't work in objc, we can start by replacing `EZConfiguration` in Swift/SwiftUI with `Configuration.swift` and make sure that the new settings page functionality works fine. `EZConfiguration` in objc can be left unchanged for now.

I don't think we really need a `Configuration` in Swift since we can access the defaults via `Defaults[.defaultKey]`. Maybe we can just leave `NewConfiguration.swift` to be the only observer for the user defaults globally (which means just delete codes like `logSetting` in `setSomething` in `EZConfiguration.m`).

tisfeng:
I thought about it carefully and found that it would be inconvenient to maintain two Configurations of objc and Swift. I would rather replace it with a Swift Configuration and not change the code on the objc side in the future.

Because the project is now using settings from objc EZConfiguration in a lot of places, there's no way to control those settings via the new settings page if we manage them separately.

> I don't think we really need a Configuration in Swift since we can access the defaults via Defaults[.defaultKey]

@CanglongCl  I know this, the properties in Configuration are for the convenience of objc code to read and write, if there is a better solution here, I am happy to accept.

tisfeng:
> @tisfeng The property in the swift configuration are using Struct. this is not accessible by the objc class.

@CanglongCl  Do we have any good ideas on how to fix this?  I'm not familiar with bridging Swift and objc.

CanglongCl:
> I thought about it carefully and found that it would be inconvenient to maintain two Configurations of objc and Swift. I would rather replace it with a Swift Configuration and not change the code on the objc side in the future.
> 
> Because the project is now using settings from objc EZConfiguration in a lot of places, there's no way to control those settings via the new settings page if we manage them separately.
> 
> > I don't think we really need a Configuration in Swift since we can access the defaults via Defaults[.defaultKey]
> 
> @CanglongCl I know this, the properties in Configuration are for the convenience of objc code to read and write, if there is a better solution here, I am happy to accept.

You are right. Currently the swift `Configuration` is good enough. Maybe the next step is to replace `EZConfiguration` in objc parts.

CanglongCl:
> > @tisfeng The property in the swift configuration are using Struct. this is not accessible by the objc class.
> 
> @CanglongCl Do we have any good ideas on how to fix this? I'm not familiar with bridging Swift and objc.

I'm not familiar to it as well. Maybe you can ask someone else. 

@NeverAgain11 BTW I'm not quite sure what did you mean as well. It seems like all properties in `Configuration` are `class` definition.

NeverAgain11:
> > > @tisfeng The property in the swift configuration are using Struct. this is not accessible by the objc class.
> > 
> > 
> > @CanglongCl Do we have any good ideas on how to fix this? I'm not familiar with bridging Swift and objc.
> 
> I'm not familiar to it as well. Maybe you can ask someone else.
> 
> @NeverAgain11 BTW I'm not quite sure what did you mean as well. It seems like all properties in `Configuration` are `class` definition.

Sorry, I didn't test it right way, the Configuration property can be accessible by the objc class.
I will replace EZConfiguration with Configuration.
@tisfeng @CanglongCl

CanglongCl:
image
`Configuration` crashed the app in SwiftUI `NewApp` (not sure if it works in old app).

tisfeng (changes requested):

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,566 @@
...
+        }.tieToLifetime(of: self)
+
+        
+        Defaults.observe(.appearanceType) { [unowned self] change in
+            let newValue = change.newValue
+            let oldValue = change.oldValue
+            print("update appearanceType: \(oldValue.title) -> \(newValue.title)")
+
+            if newValue != oldValue {
+                didSetAppearance()
```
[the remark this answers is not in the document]
NeverAgain11: It's only been called once on my side.

tisfeng:
Run error

image

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,565 @@
...
+        }.tieToLifetime(of: self)
+
+        Defaults.observe(.fontSizeOptionIndex) { [unowned self] _ in
+            didSetFontSizeIndex()
+        }.tieToLifetime(of: self)
+
+        Defaults.observe(.appearanceType) { [unowned self] change in
+            let newValue = change.newValue
+            let oldValue = change.oldValue
+            print("update appearanceType: \(oldValue.title) -> \(newValue.title)")
```
tisfeng, line 281: Actually, what I mean is the `Defaults.observe` function will be called twice, and print "update appearanceType:" twice, not just `.appearanceType`, all of the other keys are like this.

It's strange, maybe `Defaults` uses it in a wrong way somewhere, but I don't know it.

tisfeng (changes requested):

Easydict/Feature/Configuration/NewConfiguration.swift
tisfeng: When I switch from dark mode to follow system

```
update appearanceType: 深色模式 -> 跟随系统
深色模式
update appearanceType: 深色模式 -> 跟随系统
深色模式
```

https://github.com/tisfeng/Easydict/assets/25194972/f9dd25b6-9194-41b8-9fbe-2b67368f4234

tisfeng: @CanglongCl Can you take a look at this issue? We need to merge this PR as early as possible.

CanglongCl:
@tisfeng I've resolved the issue. It seems like a issue cause by `UserDefaults` instead of `Defaults` package. I found some information [here](https://stackoverflow.com/questions/65965666/why-does-user-defaults-publisher-trigger-multiple-times) on Stack OverFlow. 

Now I replaced the old code with 
```swift
Defaults.publisher(.hideMenuBarIcon)
                .removeDuplicates()
                .sink { [weak self] _ in
                    self?.didSetHideMenuBarIcon()
                }
```
where `.removeDuplicates()` should handle the problem. However, now only the changes of the value will trigger it instead of each set operation. I think it is not a problem here. 

cc @NeverAgain11 Please review the changes.

CanglongCl (approved):

NeverAgain11:
> @tisfeng I've resolved the issue. It seems like an issue cause by `UserDefaults` instead of `Defaults` package. I found some information [here](https://stackoverflow.com/questions/65965666/why-does-user-defaults-publisher-trigger-multiple-times) on Stack OverFlow.
> 
> Now I replaced the old code with
> 
> ```swift
> Defaults.publisher(.hideMenuBarIcon)
>                 .removeDuplicates()
>                 .sink { [weak self] _ in
>                     self?.didSetHideMenuBarIcon()
>                 }
> ```
> 
> where `.removeDuplicates()` should handle the problem. However, now only the changes of the value will trigger it instead of each set operation. I think it is not a problem here.
> 
> cc @NeverAgain11 Please review the changes.

It's works fine.
The DefaultsWrapper easily causes the Simultaneous Accesses error, so I change the type from struct to class.

tisfeng:
Can we remove `#import "EZConfiguration.h"` from `Easydict-Bridging-Header.h` now ?

https://github.com/tisfeng/Easydict/blob/2beb94ed7c273070c2bcd5b5c430e9da608311f1/Easydict/App/Easydict-Bridging-Header.h#L23

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,699 @@
...
+    var autoCopyOCRText: Bool
+
+    @DefaultsWrapper(.autoCopyFirstTranslatedText)
+    var autoCopyFirstTranslatedText: Bool
+
+    @DefaultsWrapper(.languageDetectOptimize)
+    var languageDetectOptimize: EZLanguageDetectOptimize
+
+//    @DefaultsWrapper(.defaultTTSServiceType)
+//    var defaultTTSServiceType: TTSServiceType
```
tisfeng, line 95: Please check and remove unused code.

Easydict/Feature/Configuration/NewConfiguration.swift
```diff
@@ -0,0 +1,699 @@
...
+
+        cancellables.append(
+            Defaults.publisher(.appearanceType)
+                .removeDuplicates()
+                .sink { [weak self] change in
+                    let newValue = change.newValue
+                    let oldValue = change.oldValue
+                    print("update appearanceType: \(oldValue.title) -> \(newValue.title)")
+
+                    if newValue != oldValue {
```
tisfeng, line 414: This code should not be needed, as well as the print above.

Easydict/Feature/Service/Baidu/EZBaiduTranslate.m
```diff
@@ -124,15 +125,16 @@ - (EZServiceType)serviceType {
 
 - (EZQueryTextType)queryTextType {
     EZQueryTextType defaultType = EZQueryTextTypeDictionary | EZQueryTextTypeSentence | EZQueryTextTypeTranslation;
-    EZQueryTextType type = [EZConfiguration.shared queryTextTypeForServiceType:self.serviceType];
+//    EZQueryTextType type = [Configuration.shared queryTextTypeForServiceType:self.serviceType];
+    EZQueryTextType type = [Configuration.shared queryTextTypeForServiceType:EZServiceTypeGoogle];
```
tisfeng, line 129: Why use `EZServiceTypeGoogle` here?

tisfeng (changes requested):

tisfeng:
> In addition, I noticed that both `NewConfiguration.swift` and `Configuration.swift` exist, maybe we need to merge them and move the Swift code all under a new group, moving to the `Configuration.swift` file.

Again, can we merge `NewConfiguration.swift` and `Configuration.swift`? They have similar names and functions, which can be confusing.

Easydict/Feature/Service/Baidu/EZBaiduTranslate.m
NeverAgain11: It's my fault.

Easydict/Feature/Service/Baidu/EZBaiduTranslate.m
```diff
@@ -124,15 +125,17 @@ - (EZServiceType)serviceType {
 
 - (EZQueryTextType)queryTextType {
     EZQueryTextType defaultType = EZQueryTextTypeDictionary | EZQueryTextTypeSentence | EZQueryTextTypeTranslation;
-    EZQueryTextType type = [EZConfiguration.shared queryTextTypeForServiceType:self.serviceType];
+//    EZQueryTextType type = [Configuration.shared queryTextTypeForServiceType:self.serviceType];
```
tisfeng, line 128:
```suggestion
```

tisfeng (changes requested):

CanglongCl (approved):

AkaShark (approved):

phlpsong (approved):

Jerry23011 (approved):

tisfeng (approved):

tisfeng:
Thank you all, the new settings page is mostly working now, with probably a few minor issues, hopefully we can finalize and polish it up soon and get it out to users.

Easydict/Feature/Configuration/Configuration.swift
```diff
@@ -0,0 +1,701 @@
...
+    }
+
+    func intelligentQueryModeForWindowType(_ windowType: EZWindowType) -> Bool {
+        let key = EZConstKey.constkey("IntelligentQueryMode", windowType: windowType)
+        let defaultValue = "0"
+        // Turn on intelligent query mode by default in mini window.
+        if windowType == .mini {
+            return true
+        }
+        return UserDefaults.standard.string(forKey: key) ?? defaultValue == "1"
```
tisfeng, line 651: 上面代码有问题，会导致 mini 窗口一直开启智能查询模式，且无法手动关闭 https://github.com/tisfeng/Easydict/issues/502#issuecomment-2094206165 。

以下是之前功能的 objc 代码。
```objc
- (BOOL)intelligentQueryModeForWindowType:(EZWindowType)windowType {
    NSString *key = [EZConstKey constkey:EZIntelligentQueryModeKey windowType:windowType];
    
    NSString *defaultValue = @"0";
    // Turn on intelligent query mode by default in mini window.
    if (windowType == EZWindowTypeMini) {
        defaultValue = @"1";
    }
    
    NSString *stringValue = [NSUserDefaults mm_readString:key defaultValue:defaultValue];
    return [stringValue boolValue];
}
```

PR #340, opened 2024-01-19
add disabled app list tab

phlpsong:
close https://github.com/tisfeng/Easydict/issues/289

After a lot of research, I found that the selection highlight style could not customized(with multi selection gesture support) due to SwiftUI's limitation. 

Any questions or suggestions about this, please let me know.


Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -0,0 +1,225 @@
...
+            .fileImporter(
+                isPresented: $disabledAppViewModel.isImporting,
+                allowedContentTypes: [.application],
+                allowsMultipleSelection: true
+            ) { result in
+                switch result {
+                case let .success(urls):
+                    disabledAppViewModel.newAppURLsSelected(from: urls)
+                case let .failure(error):
+                    print("fileImporter error: \(error)")
```
CanglongCl, line 73: it is recommended to pop an alert to show there is an error to import the file.

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -0,0 +1,225 @@
...
+        .frame(maxWidth: 500)
+        .environmentObject(disabledAppViewModel)
+        .onAppear {
+            disabledAppViewModel.fetchDisabledApps()
+        }
+    }
+}
+
+@available(macOS 13.0, *)
+struct ListToolbar: View {
```
CanglongCl, line 120: Those views that is only used by `DisabledAppTab` should be marked as `private`.

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -0,0 +1,225 @@
...
+        .padding(.vertical, 4)
+        .padding(.leading, 6)
+        .task {
+            appFetcher.getAppBundleInfo()
+        }
+    }
+}
+
+@available(macOS 13.0, *)
+class AppFetcher: ObservableObject {
```
CanglongCl, line 196: I'm not sure if we really need a observable object here. Maybe we can just nested `getAppBundleInfo()` inside the initialization of `BlockAppItemView`?

Jerry23011:
I think it looks very nice, finally made it through all the tabs in Settings 🎉

Just some notes here:
- My system is in Simplified Chinese, why is Keychain Access.app displayed in English
截屏2024-01-18 19 03 39

-  There seem to be an almost unnoticeable lag for the list to appear when switching from Service tab to Disabled app list tab

https://github.com/tisfeng/Easydict/assets/89069957/829d16a9-d199-4500-9c4f-9d231e62e949

CanglongCl (changes requested):

CanglongCl:
> close #289
> 
> After a lot of research, I found that the selection highlight style could not customized(with multi selection gesture support) due to SwiftUI's limitation.
> 
> Any questions or suggestions about this, please let me know.

There's indeed no easy solution for it if you are using `List(selection)`. If it need more customization, use `List()` and detect with `selectedItems.contain(item)` instead

phlpsong:
> I think it looks very nice, finally made it through all the tabs in Settings 🎉
> 
> Just some notes here:
> 
> * My system is in Simplified Chinese, why is Keychain Access.app displayed in English
> 
> 截屏2024-01-18 19 03 39
> * There seem to be an almost unnoticeable lag for the list to appear when switching from Service tab to Disabled app list tab
> 
>  2024-01-18.19.04.04.mov

Thanks for your comments. 

First issue fixed in [502a346](https://github.com/tisfeng/Easydict/pull/340/commits/502a3468be5645fdcdd607232fa4655a8df529a5)

About the second issue I will take a look at it later. The odd behavior happens when service tab switch to others tab, I think root cause is the frame size change.

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -0,0 +1,225 @@
...
+        VStack {
+            Text("disabled_title")
+                .padding()
+
+            appListViewWithToolbar
+        }
+        .frame(maxWidth: 500)
+        .environmentObject(disabledAppViewModel)
+        .onAppear {
+            disabledAppViewModel.fetchDisabledApps()
```
CanglongCl, line 114: We should call it while initializing the `DisabledAppViewModel` instead of view appearing

phlpsong:
> > close #289
> > After a lot of research, I found that the selection highlight style could not customized(with multi selection gesture support) due to SwiftUI's limitation.
> > Any questions or suggestions about this, please let me know.
> 
> There's indeed no easy solution for it if you are using `List(selection)`. If it need more customization, use `List()` and detect with `selectedItems.contain(item)` instead

Yes, I have thought about this before, but we will lose the quick multi selection ability after that.  Maybe we could discuss and decide which one is more important.

CanglongCl (changes requested):

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
phlpsong: ok, done.

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
phlpsong: Yes, removed in [ceb5e36](https://github.com/tisfeng/Easydict/pull/340/commits/ceb5e3648b6b111dc26ffa636544dcd86c74a003),  looks more clear.

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -162,47 +169,34 @@ struct ListButton: View {
...
 
             Spacer()
         }
         .frame(maxWidth: .infinity)
         .contentShape(Rectangle())
         .padding(.vertical, 4)
         .padding(.leading, 6)
-        .task {
-            appFetcher.getAppBundleInfo()
+        .onAppear {
```
CanglongCl, line 196: Again, I think we don't need a `onAppear` here. I'll prefer 
```swift
let appIcon: NSImage
let appName: String

init(app: EZAppModel) {
    self.appIcon = ...
    self.appName = ...
}
```

How do you think?

CanglongCl (approved):

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
phlpsong: `State` value could not update in `init`, so I put it in here.

If I understand right, you mean put above code in `BlockAppItemView`, but it will need decorate them as `State` or we could not change them in struct.

CanglongCl: Yes. I don't think we need change `appIcon` and `appName` so they don't need to be decorated with `@State`. Just put the entire logic of `getAppBundleInfo()` inside `init()`

phlpsong: ok, it makes sense to me.

CanglongCl: Great~

phlpsong:
@Jerry23011 The lag issue disappears in the latest commit. You may double check it.

Jerry23011:
Yes indeed, both of them are gone. Thank you

Jerry23011 (dismissed):

CanglongCl (approved):

tisfeng:
It looks great, but there are two small issues that need to be fixed:

1. Opening the system control and adding a disabled App by dragging it causes a very noticeable lag.
2. As for adding and removing disabled App controls, I prefer the previous segmented style, it's more obvious and clearer.

In addition, the current Add and Remove buttons do not have clear enough colors, there should be some contrast directly between the enabled and disabled states of the buttons, which needs to be improved by referring to the previous buttons.

<table>
    <td> 
    <td> 
</table>

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -0,0 +1,224 @@
...
+@available(macOS 13.0, *)
+private struct ListToolbar: View {
+    @EnvironmentObject private var disabledAppViewModel: DisabledAppViewModel
+
+    var body: some View {
+        VStack(spacing: 0) {
+            Divider()
+            HStack(spacing: 0) {
+                ListButton(imageName: "plus") {
+                    disabledAppViewModel.isImporting.toggle()
```
tisfeng, line 135: I had this drag causing lag problem before and solved it in a very dumb way, maybe you can optimize this ugly code.

```objc
EZConfiguration.shared.disabledAutoSelect = YES;
...
add a disabled App 
...
EZConfiguration.shared.disabledAutoSelect = NO;
```

https://github.com/tisfeng/Easydict/blob/2df80b922865103b24d19a22b541b89c3218bf3e/Easydict/Feature/PerferenceWindow/DisableAutoSelectTextViewController/EZDisableAutoSelectTextViewController.m#L257

tisfeng (changes requested):

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
phlpsong: Yes, I think this is the reason of first issue.

CanglongCl:
> > close #289
> > After a lot of research, I found that the selection highlight style could not customized(with multi selection gesture support) due to SwiftUI's limitation.
> > Any questions or suggestions about this, please let me know.
> 
> Customizing the highlight background color is not necessary, I just wanted to keep it consistent with the old settings page UI, but then I realized that the overall color of the new settings page has changed, and it looks fine with the system default highlight color.
> 
> I also didn't realize that customizing the highlight color in SwiftUI would be such a troublesome thing, and that it would cause onMove events and multi-select events to fail 😓

We can change the accent color for components like button and toggle with changing `accentColor` in assets

phlpsong:
> It looks great, but there are two small issues that need to be fixed:
> 
> 
> 
> 1. Opening the system control and adding a disabled App by dragging it causes a very noticeable lag.
> 
> 2. As for adding and removing disabled App controls, I prefer the previous segmented style, it's more obvious and clearer.
> 
> 
> 
> In addition, the current Add and Remove buttons do not have clear enough colors, there should be some contrast directly between the enabled and disabled states of the buttons, which needs to be improved by referring to the previous buttons.
> 
> 
> 
> <table>
> 
>     <td> 
> 
>     <td> 
> 
> </table>
> 
> 

I did tried to build a picker style component but could not make the appearance perfect, so I refer the system list component to build this one.

tisfeng:
Using the system style is fine, But this background color needs to be modified to distinguish it from the previous cell, as well as the display color of the button `+` and `-`.

image

phlpsong:
> > > close #289
> 
> > > After a lot of research, I found that the selection highlight style could not customized(with multi selection gesture support) due to SwiftUI's limitation.
> 
> > > Any questions or suggestions about this, please let me know.
> 
> > 
> 
> > Customizing the highlight background color is not necessary, I just wanted to keep it consistent with the old settings page UI, but then I realized that the overall color of the new settings page has changed, and it looks fine with the system default highlight color.
> 
> > 
> 
> > I also didn't realize that customizing the highlight color in SwiftUI would be such a troublesome thing, and that it would cause onMove events and multi-select events to fail 😓
> 
> 
> 
> We can change the accent color for components like button and toggle with changing `accentColor` in assets

I all solutions I ever seen, included 'accentColor' modifier which will deprecated later and '.tint', also add 'accent' color in color set,but neither of them works in list case.

We customized this highlight behavior by override 'NSTableRowView' in AppKit which not available in SwiftUI.

phlpsong:
> Using the system style is fine, But this background color needs to be modified to distinguish it from the previous cell, as well as the display color of the button `+` and `-`.
> 
> 
> 
> image
> 
> 

OK, will update this later

phlpsong:
> > Using the system style is fine, But this background color needs to be modified to distinguish it from the previous cell, as well as the display color of the button `+` and `-`.
> > image
> 
> OK, will update this later

I tried but could not find a suitable color for the list background. This must use the system form background color or something else.

tisfeng:
If there is no suitable system color, we can add one ourselves, as long as it achieves the effect we want.

I referred to the system settings page and added an `add & minus` background color and list border color,  it looks good.

image

phlpsong:
@tisfeng  yep, I did tried add the same color with system table bg, but not look harmonized. 

I think your changes look good.

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -118,11 +115,11 @@ struct DisabledAppTab: View {
     var body: some View {
         VStack {
             Text("disabled_title")
-                .padding()
+                .padding(.top, 18)
+                .padding(.bottom, 8)
```
phlpsong, line 119: This will change text horizontal padding. 

Screenshot 2024-01-20 at 08 43 43

tisfeng: ok

tisfeng (approved):

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -0,0 +1,230 @@
...
+        }
+        .frame(height: 28)
+        .background(Color("add_minus_bg_color"))
+    }
+}
+
+@available(macOS 13.0, *)
+private struct ListButton: View {
+    var imageName: String
+    var enabled: Bool
```
CanglongCl, line 157: We can use `@Environment(\.isEnabled) var enabled` to replace explicitly passing a `enabled` into the view.

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -0,0 +1,230 @@
...
+            .padding(2)
+        }
+        .frame(height: 28)
+        .background(Color("add_minus_bg_color"))
+    }
+}
+
+@available(macOS 13.0, *)
+private struct ListButton: View {
+    var imageName: String
```
CanglongCl, line 156: Better named as `systemName` just like the `Image`'s parameter, otherwise it may be confusing.

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -0,0 +1,230 @@
...
+//  Easydict
+//
+//  Created by phlpsong on 2024/1/15.
+//  Copyright © 2024 izual. All rights reserved.
+//
+
+import Combine
+import SwiftUI
+
+class DisabledAppViewModel: ObservableObject {
```
CanglongCl, line 12: I think we can mark the view model as `private`

CanglongCl (changes requested):

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
phlpsong: sure.

phlpsong:
@tisfeng  Could you describe the detail of the lag issue? Or we can merge this now and open an issue to track it separately?

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
```diff
@@ -133,15 +133,16 @@ private struct ListToolbar: View {
...
-                let isNoSelectedApps = disabledAppViewModel.selectedAppModels.isEmpty
-                ListButton(imageName: "minus", enabled: !isNoSelectedApps) {
+                .environment(\.isEnabled, true)
+                Divider()
+                    .padding(.vertical, 1)
+                ListButton(systemName: "minus") {
                     disabledAppViewModel.removeDisabledApp()
                 }
-                .disabled(isNoSelectedApps)
+                .environment(\.isEnabled, !disabledAppViewModel.selectedAppModels.isEmpty)
```
CanglongCl, line 145: Ouch, here we should use `.disabled` instead of `.environment`.

CanglongCl (changes requested):

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
phlpsong: Nice, lesson learned.

CanglongCl (dismissed):

Easydict/NewApp/View/SettingView/Tabs/DisabledAppTab.swift
tisfeng: Ok, I will open a new issue to track it.

tisfeng (approved):
Looks good, let's merge it.

Jerry23011 (approved):

PR #341, opened 2024-01-19
perf: adjust setting UI

tisfeng:
Keep the settings page windows all the same width to avoid strange animations.

Refer to  [PasteNow](https://pastenow.app/)'s settings page animation.

## Before

https://github.com/tisfeng/Easydict/assets/25194972/5f5d1041-7472-4c6d-9432-634a1045e483

## After

https://github.com/tisfeng/Easydict/assets/25194972/9fd57bd4-365d-4018-8d5f-90398a4ef723

### DIFF
--- a/Easydict/NewApp/View/SettingView/SettingView.swift
+++ b/Easydict/NewApp/View/SettingView/SettingView.swift
@@ -52,14 +52,22 @@ struct SettingView: View {
     func resizeWindowFrame() {
         guard let window else { return }
 
-        let originalFrame = window.frame
-        let newSize = switch selection {
-        case .general, .privacy, .about:
-            CGSize(width: 500, height: 520)
+        // Keep the settings page windows all the same width to avoid strange animations.
+        let maxWidth = 650
+        let height = switch selection {
+        case .general:
+            maxWidth
         case .service:
-            CGSize(width: 800, height: 520)
+            500
+        case .privacy:
+            320
+        case .about:
+            450
         }
 
+        let newSize = CGSize(width: maxWidth, height: height)
+
+        let originalFrame = window.frame
         let newY = originalFrame.origin.y + originalFrame.size.height - newSize.height
         let newRect = NSRect(origin: CGPoint(x: originalFrame.origin.x, y: newY), size: newSize)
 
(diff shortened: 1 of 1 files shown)


tisfeng:
The layout of About page does not look good at the moment, it might need some adjustment later, but I don't have a better idea yet.

image

image

CanglongCl (approved):

Jerry23011:
I'm wondering if we could dim the maximize button in the settings window, as noted in Apple's HID standards:
> Dim a settings window’s minimize and maximize buttons. It’s quick to open a settings window using the standard Command–Comma (,) keyboard shortcut, so there’s no need to keep the window in the Dock, and because a settings window accommodates the size of the current pane, people don’t need to expand the window to see more.

CC @tisfeng @CanglongCl

tisfeng:
OK, we can disable the zoom button.

Easydict/NewApp/View/SettingView/SettingView.swift
```diff
@@ -51,15 +51,26 @@ struct SettingView: View {
 
     func resizeWindowFrame() {
         guard let window else { return }
+        
+        // Disable zoom button, ref: https://stackoverflow.com/a/66039864/8378840
+        window.standardWindowButton(.zoomButton)?.isEnabled = false
```
tisfeng, line 56: It seems that disabling the zoom button in SwiftUI is not easy, I found a workaround from here  https://stackoverflow.com/a/66039864/8378840

Please let me know if there is a better solution.

image

CanglongCl (approved):

Easydict/NewApp/View/SettingView/SettingView.swift
CanglongCl: Looks good.

Jerry23011 (approved):

phlpsong:
Maybe we can move `Check for Updates` to the general tab. 


This one looks great too.
Screenshot 2024-01-20 at 15 16 48

tisfeng:
> Maybe we can move Check for Updates to the general tab.

Ok, I'm not good at product design, suggestions are welcome 🤗

Jerry23011:
> Maybe we can move Check for Updates to the general tab.

+1

CanglongCl:
> > Maybe we can move Check for Updates to the general tab.
> 
> Ok, I'm not good at product design, suggestions are welcome 🤗

I'll prefer put it in general tab just like more apps. It is more consistent with the user's mental model.

tisfeng:
It looks like everyone is in agreement, so let's move it to the general page later.

Issue #346, opened 2024-01-20
dev: Adding a disabled App by dragging it causes a very noticeable lag


tisfeng:
From  https://github.com/tisfeng/Easydict/pull/340#discussion_r1458604586

https://github.com/tisfeng/Easydict/assets/25194972/09958504-2f96-4939-9ff2-c8d0100ab92d

tisfeng:
The problem may be related to get select word by mouse, I didn't find the exact reason, here is my previous fix code 0b8f9e9

phlpsong:
ok, I will check it

phlpsong:
I tested it and could not reproduce it on my 2 Mac devices(M2 chip model & Intel chip 2018 model).

[screen-capture.webm](https://github.com/tisfeng/Easydict/assets/103433299/296b9d79-0f9e-43f8-8823-60b3f9083897)

I wonder if any others have this issue and do you have some special settings in your side?

tisfeng:
I tested this code, it doesn't fix on my mac.

Maybe I forgot to say,  option  `auto select text` should be turned on.

<img width="617" alt="image" src="https://github.com/tisfeng/Easydict/assets/25194972/cc11542f-4ae5-40af-9aa2-a861f50399b3">

PR #355, opened 2024-01-23
fix add disable app drag lag issue

phlpsong:
close [#346](https://github.com/tisfeng/Easydict/issues/346)

### DIFF
--- a/Easydict/Feature/EventMonitor/EZEventMonitor.m
+++ b/Easydict/Feature/EventMonitor/EZEventMonitor.m
@@ -143,8 +143,15 @@ static EZEventMonitor *_instance = nil;
     wechat.appBundleID = @"com.tencent.xinWeChat";
     wechat.triggerType = EZTriggerTypeDoubleClick | EZTriggerTypeTripleClick;
     
+    // https://github.com/tisfeng/Easydict/issues/346
+    // Add Easydict self as block app
+    EZAppModel *selfApp = [[EZAppModel alloc] init];
+    selfApp.appBundleID = [[NSBundle mainBundle] bundleIdentifier];
+    selfApp.triggerType = EZTriggerTypeNone;
+    
     NSArray *defaultAppModels = @[
         wechat,
+        selfApp
     ];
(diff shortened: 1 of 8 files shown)


Easydict/Feature/EventMonitor/EZEventMonitor.m
```diff
@@ -143,8 +143,15 @@ - (EZTriggerType)getAppSelectTextActionType:(NSString *)appBundleID
     wechat.appBundleID = @"com.tencent.xinWeChat";
     wechat.triggerType = EZTriggerTypeDoubleClick | EZTriggerTypeTripleClick;
     
+    // https://github.com/tisfeng/Easydict/issues/346
+    // Add Easydict self as block app
+    EZAppModel *selfApp = [[EZAppModel alloc] init];
+    selfApp.appBundleID = [[NSBundle mainBundle] bundleIdentifier];
+    selfApp.triggerType = EZTriggerTypeNone;
```
tisfeng, line 150: No, that's not a good solution, and although we're not currently handling the mouse event inside Easydict, it may be used later, such as in a future main window, so we can't just turn it off here.

If there is no better solution, you can use the old workaround https://github.com/tisfeng/Easydict/commit/0b8f9e9b0b5932ea95ce91258dc114243011b82c for now.

tisfeng (changes requested):

Easydict/Feature/EventMonitor/EZEventMonitor.m
phlpsong: ok, I will use the old workaround for now. I'll see if I can find other solutions.

tisfeng (approved):

tisfeng:
Note that PR requires the approval of more than **two** reviewers to merge.

AkaShark (approved):

phlpsong:
> Note that PR requires the approval of more than **two** reviewers to merge.

OK

Jerry23011 (approved):

Issue #344, opened 2024-01-20
【UI/UX】Design Issues Tracker for Swift/SwiftUI migration


Jerry23011:
This issue is used to track design-related issues in PRs that migrates Easydict to Swift/SwiftUI

- [x] Add copyright info to Settings, About tab
#309

- [x] In Settings, the button appears to experience lag when transitioning from other tabs to General
#326 

- [ ] ~~Scrolling in Settings, General tab feels laggy on 120Hz display~~ (SwiftUI performance issue)
#326 

- [x] Settings window often fails to appear in the front
#326

- [x] Add alert when user tries to hide menubar icon
#345 

- [ ] ~~Move `Check for Update` button to Settings, Advanced Tab~~
https://github.com/tisfeng/Easydict/pull/341#issuecomment-1901924275

- [ ] ~~Making tabs in Settings scrollable or not (General/Advanced/Privacy)~~
#359 
- [x] Some strings in Settings, Service tab are not localized
#311 
#324 
- [x] In Settings, Services tab, the list seems to wear off along with the elements when scrolling to the top/bottom
#326

Jerry23011:
Closing this as most of the issues are resolved or consensus is reached.

PR #359, opened 2024-01-24
Disable scrolling in Privacy Tab

Jerry23011:
We will try to make each tab “unscrollable" so that it's consistent with other macOS app settings.

### DIFF
--- a/Easydict/NewApp/View/SettingView/Tabs/PrivacyTab.swift
+++ b/Easydict/NewApp/View/SettingView/Tabs/PrivacyTab.swift
@@ -31,6 +31,7 @@ struct PrivacyTab: View {
             }
         }
         .formStyle(.grouped)
+        .scrollDisabled(true)
     }
 
     @AppStorage("EZConfiguration_kAllowCrashLogKey")


CanglongCl (approved):

phlpsong (approved):

tisfeng (changes requested):
About the Privacy Tab, I don't have a strong feeling on whether to make it scrollable or not.

I had a look at the macOS system settings, basically all of them are scrollable, even for the ones that have very few elements, such as the Software Update pane, which I guess is also implemented with SwiftUI's `Form`.

So I prefer to keep it unchanged, let it be scrollable by default. 

image

Jerry23011:
> I had a look at the macOS system settings, basically all of them are scrollable, even for the ones that have very few elements, such as the Software Update pane, which I guess is also implemented with SwiftUI's `Form`.

I would say that we are comparing with other macOS apps instead of the system setting. The settings views are not scrollable in Apple’s default apps, so I’m thinking we want to follow that convention

CanglongCl:
I don't think we can make all setting view fixed since our setting is so heavy and will get heavier in the future. For privacy tab, it is not a big problem since it is hardly perceived it.

Jerry23011:
> I don't think we can make all setting views fixed since our settings are so heavy and will get heavier in the future. 

That's a very good point. Perhaps we can leave this PR on hold for a while as we observe how the migration goes.

What do you think @tisfeng

Jerry23011:
Wait, I just thought of a use case. Since users are scrolling through General/Advanced; it's natural that they think Privacy is scrollable.

Making Privacy unscrollable would create inconsistencies.

Closing this, will track it in #344

tisfeng:
Yes, That is also an aspect I have considered.

PR #360, opened 2024-01-24
Migrate to Defaults for Privacy & About tabs

Jerry23011:


CanglongCl (approved):

tisfeng (approved):

Easydict/NewApp/View/SettingView/Tabs/AboutTab.swift
```diff
@@ -45,8 +46,7 @@ struct AboutTab: View {
         }
     }
 
-    @AppStorage("EZConfiguration_kAutomaticallyChecksForUpdatesKey")
-    private var autoChecksForUpdates = false
+    @Default(.autoCheckForUpdates) private var autoChecksForUpdates
```
tisfeng, line 49: Maybe we do not need to save `autoCheckForUpdates` to defaults.

Refer Sparkle docs:

```objc
/**
 A property indicating whether or not to check for updates automatically.
 
 By default, Sparkle asks users on second launch for permission if they want automatic update checks enabled
 and sets this property based on their response. If `SUEnableAutomaticChecks` is set in the Info.plist,
 this permission request is not performed however.
 
 Setting this property will persist in the host bundle's user defaults.
 Hence developers shouldn't maintain an additional user default for this property.
 Only set this property if the user wants to change the default via a user settings option.
 Do not always set it on launch unless you want to ignore the user's preference.
 For testing environments, you can disable update checks by passing `-SUEnableAutomaticChecks NO`
 to your app's command line arguments instead of setting this property.
 
 The update schedule cycle will be reset in a short delay after the property's new value is set.
 This is to allow reverting this property without kicking off a schedule change immediately
 */
@property (nonatomic) BOOL automaticallyChecksForUpdates;
```

tisfeng: How about set `SUEnableAutomaticChecks` to YES in the Info.plist?

This is old code:

https://github.com/tisfeng/Easydict/blob/2df80b922865103b24d19a22b541b89c3218bf3e/Easydict/Feature/Configuration/EZConfiguration.m#L146

tisfeng (changes requested):

Easydict/NewApp/View/SettingView/Tabs/AboutTab.swift
CanglongCl: @Jerry23011 Try binding 
```swift
var autoChecksForUpdates: Binding<Bool> {
    .init(get:set)
}
```

AkaShark (approved):

tisfeng (approved):

tisfeng:
> Waiting for #363

#363 has fixed, please continue.

Easydict/NewApp/View/SettingView/Tabs/AboutTab.swift
tisfeng: This code doesn't work, please check it.

CanglongCl: Fixed with some ugly codes in 28c9c02. Any suggestion?

tisfeng: It works, we can use this code if there is no better solution.

Jerry23011: Thanks @CanglongCl

tisfeng:
Please review this PR.

phlpsong (dismissed):

CanglongCl (approved):

tisfeng (approved):

Jerry23011:
thanks, everyone

PR #361, opened 2024-01-24
fix: General tab other section title

phlpsong:


Jerry23011 (dismissed):
We could think twice if we want the gender neutral pronoun “其它“. Apple’s official Simplified Chinese localization adopts “其他”. LGTM never the less.

phlpsong:
> We could think twice if we want the gender neutral pronoun “其它“. Apple’s official Simplified Chinese localization adopts “其他”. LGTM never the less.

Good point! I checked some articles, `其他` is more accurate.

CanglongCl (approved):

Jerry23011 (approved):

AkaShark (approved):

PR #364, opened 2024-01-25
fix: setting crash after uninstall app in disabled app list

phlpsong:
Settings could not open after uninstall an app in the disabled app list in AppKit mode, SwiftUI version will show empty content

reproduce steps:
1. add one app in the disabled app list
3. uninstall the app
4. open settings

### DIFF
--- a/Easydict/Feature/PerferenceWindow/DisableAutoSelectTextViewController/EZDisableAutoSelectTextViewController.m
+++ b/Easydict/Feature/PerferenceWindow/DisableAutoSelectTextViewController/EZDisableAutoSelectTextViewController.m
@@ -57,7 +57,7 @@ static NSString *const EZColumnId = @"EZColumnId";
 }
 
 - (void)setup {
-    self.appModelList = [[EZLocalStorage.shared selectTextTypeAppModelList] mutableCopy];
+    [self setupAppModelList];
     
     [self.titleTextField mas_makeConstraints:^(MASConstraintMaker *make) {
         make.top.left.right.inset(kMargin + 5); // ???: Why is the actual inset is 18?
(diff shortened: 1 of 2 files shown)


Easydict/Feature/PerferenceWindow/DisableAutoSelectTextViewController/EZDisableAutoSelectTextViewController.m
```diff
@@ -78,6 +78,18 @@ - (void)setup {
...
 
+- (void)setupAppModelList {
+    self.appModelList = [[NSMutableArray alloc] init];
+    NSArray<EZAppModel *> *allAppModelList = [EZLocalStorage.shared selectTextTypeAppModelList];
+    NSWorkspace* workspace = [NSWorkspace sharedWorkspace];
+    for (EZAppModel *appModel in allAppModelList) {
+        NSURL *appURL = [workspace URLForApplicationWithBundleIdentifier:appModel.appBundleID];
+        if (appURL) {
+            [self.appModelList addObject:appModel];
+        }
```
tisfeng, line 89: I debugged the old code, found that when uninstalling the App in the disabled list, `appURL` is nil , which indeed causes the following method to crash 😓

```objc
NSBundle *appBundle = [[NSBundle alloc] initWithURL:appURL];
```

https://github.com/tisfeng/Easydict/blob/dev/Easydict/Feature/PerferenceWindow/DisableAutoSelectTextViewController/EZAppCell.m#L65

tisfeng (approved):
Good

Easydict/Feature/PerferenceWindow/DisableAutoSelectTextViewController/EZDisableAutoSelectTextViewController.m
phlpsong: Yes, it's very corner case. 😅

CanglongCl (approved):

Issue #382, opened 2024-02-02
dev: Settings window often fails to appear in the front


tisfeng:
This issue often occurs in debug mode, which is very inconvenient for debugging. 

I'm not sure if it's a SwiftUI bug,  but we need to try to fix it.

See in point 4 https://github.com/tisfeng/Easydict/pull/326#issuecomment-1894876950

PR #383, opened 2024-02-03
fix: Setting window fails to appear front

phlpsong:
#382 

Introduce one new package: https://github.com/orchetect/SettingsAccess

This workaround does nothing if running from Xcode, but running as a standalone app it will work according to doc.

But it seems to work well both from Xcode build or archive based on my test.

### DIFF
--- a/Easydict/NewApp/View/MenuItemView.swift
+++ b/Easydict/NewApp/View/MenuItemView.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2023 izual. All rights reserved.
 //
 
+import SettingsAccess
 import Sparkle
 import SwiftUI
 import ZipArchive
@@ -24,6 +25,7 @@ final class MenuItemStore: ObservableObject {
 @available(macOS 13, *)
 struct MenuItemView: View {
     @ObservedObject private var store = MenuItemStore()
+    @Environment(\.openSettings) private var openSettings
 
     var body: some View {
         // ️.menuBarExtraStyle为 .menu 时某些控件可能会失效 ，只能显示内容（按照菜单项高度、图像以 template 方式渲染）无法交互 ，比如 Stepper、Slider 等，像基本的 Button、Text、Divider、Image 等还是能正常显示的。
@@ -92,7 +94,13 @@ struct MenuItemView: View {
     @ViewBuilder
     private var settingItem: some View {
         if #available(macOS 14.0, *) {
-            SettingsLink()
+            SettingsLink {
+                Text("Settings...")
+            } preAction: {
+                NSApp.activate(ignoringOtherApps: true)
+            } postAction: {
+                // nothing to do
+            }
         } else {
             Button("Settings...") {
                 NSLog("打开设置")
--- a/Easydict/NewApp/View/MenuItemView.swift
+++ b/Easydict/NewApp/View/MenuItemView.swift
@@ -25,7 +25,6 @@ final class MenuItemStore: ObservableObject {
 @available(macOS 13, *)
 struct MenuItemView: View {
     @ObservedObject private var store = MenuItemStore()
-    @Environment(\.openSettings) private var openSettings
 
     var body: some View {
         // ️.menuBarExtraStyle为 .menu 时某些控件可能会失效 ，只能显示内容（按照菜单项高度、图像以 template 方式渲染）无法交互 ，比如 Stepper、Slider 等，像基本的 Button、Text、Divider、Image 等还是能正常显示的。
--- a/Easydict/NewApp/View/MenuItemView.swift
+++ b/Easydict/NewApp/View/MenuItemView.swift
@@ -97,6 +97,7 @@ struct MenuItemView: View {
             SettingsLink {
                 Text("Settings...")
             } preAction: {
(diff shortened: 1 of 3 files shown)


CanglongCl:
Did it fix in macOS 13.0? It seems only affect macOS 14.0+.

phlpsong:
> Is it fixed in macOS 13.0? It seems only affect macOS 14.0+.

Yes, I thought it only affects on macOS 14.0+ using `SettingLink`. 

I think `NSApp.activate(ignoringOtherApps: true)` should work on macOS 13 too, could you have a try if you're using macOS 13?

CanglongCl:
> > Is it fixed in macOS 13.0? It seems only affect macOS 14.0+.
> 
> Yes, I thought it only affects on macOS 14.0+ using `SettingLink`.
> 
> I think `NSApp.activate(ignoringOtherApps: true)` should work on macOS 13 too, could you have a try if you're using macOS 13?

I do not use macOS 13 currently. Any other one can help?

Jerry23011:
I can setup a virtual machine later

I'll check the release mode only since it's too complicated to set up Xcode from gound up.

Jerry23011:
The issue still exists on macOS 13

I tested it on a macOS 13.6 VM

https://github.com/tisfeng/Easydict/assets/89069957/f5bda160-26f5-47b3-afa0-c5614e970748

The workaround is working on a macOS 14.4 beta 1 VM, I can see there's a workaround going, but it's pretty stable.

https://github.com/tisfeng/Easydict/assets/89069957/4bcfd8dc-c5f1-4faa-aa62-c54948d7e47c

phlpsong:
> The issue still exists on macOS 13
> 
> I tested it on a macOS 13.6 VM
> 
>  macOS.13.6.mov 
> The workaround is working on a macOS 14.4 beta 1 VM, I can see there's a workaround going, but it's pretty stable.
> 
>  macOS.14.4.beta.1.mov

I just push a new commit to apply `NSApp.activate(ignoringOtherApps: true)` below macOS 14. 

Could u give it a try?

Jerry23011:
Yep! Now it works on 13.6

Jerry23011 (dismissed):

phlpsong:
> Yep! Now it works on 13.6

Much appreciated!

CanglongCl (dismissed):

tisfeng:
I tested this code, it works.

So, is this a bug in the use of SettingsLink in SwiftUI? 🤔

Easydict/NewApp/View/MenuItemView.swift
```diff
@@ -92,10 +94,18 @@ struct MenuItemView: View {
...
+            } preAction: {
+                NSLog("打开设置")
+                NSApp.activate(ignoringOtherApps: true)
+            } postAction: {
+                // nothing to do
+            }
         } else {
             Button("Settings...") {
                 NSLog("打开设置")
+                NSApp.activate(ignoringOtherApps: true)
```
tisfeng, line 108: `activate(ignoringOtherApps:)` will be deprecated, can we use NSApp.activate instead?

https://developer.apple.com/documentation/appkit/nsapplication/1428468-activate

```swift
    /// Makes the receiver the active app.
    @available(macOS, introduced: 10.0, deprecated: 100000, message: "This method will be deprecated in a future release. Use NSApp.activate instead.")
    open func activate(ignoringOtherApps flag: Bool)
```

tisfeng (changes requested):

Easydict/NewApp/View/MenuItemView.swift
phlpsong: `NSApp.active()` is only available in macOS 14.

I tested in macOS14, and it does not work well both from Xcode build and release.

phlpsong:
> I tested this code, it works.
> 
> So, is this a bug in the use of SettingsLink in SwiftUI? 🤔

It can be said that Apple does not provide extra interfaces for `SettingsLink` to let us control the activation of window.

Easydict/NewApp/View/MenuItemView.swift
tisfeng: Ok, let's keep it.

tisfeng:
Oh, I see.

tisfeng (approved):

Easydict/NewApp/View/MenuItemView.swift
```diff
@@ -24,6 +25,7 @@ final class MenuItemStore: ObservableObject {
 @available(macOS 13, *)
 struct MenuItemView: View {
     @ObservedObject private var store = MenuItemStore()
+    @Environment(\.openSettings) private var openSettings
```
tisfeng, line 28: It seems that this variable is not used?

tisfeng (changes requested):

Easydict/NewApp/View/MenuItemView.swift
phlpsong: Yes, forgot to remove it.  Another workaround used this at first.

AkaShark (approved):

Jerry23011 (approved):

tisfeng (approved):

PR #410, opened 2024-02-16
bugfix: settings window display issue

### LINKED ISSUE #362: dev: When displaying the query window, the settings behind it is also displayed
It's a known issue that when a query window is displayed, such as the mini window, it also displays the settings page, or the main window if it's turned on, along with it.

Since I'm not familiar with macOS window management, I haven't found a good solution for this, so my previous workaround was: when displaying a query window, if the settings page is displayed, close the settings page first. If the main window is also displayed, `orderOut:`, then `orderBack:`.

This solution is ugly, but it can be used temporarily if there is no better solution.

This is old code:

https://github.com/tisfeng/Easydict/blob/2df80b922865103b24d19a22b541b89c3218bf3e/Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m#L411

https://github.com/tisfeng/Easydict/assets/25194972/d0790962-6c9e-4c64-af54-9c53c7df7e38

phlpsong:
close #362 

also fixed the issue that uses the shortcut key `cmd` + `,` to popup settings window while the query window is shown.

### DIFF
--- a/Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
+++ b/Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
@@ -413,6 +413,14 @@ static EZWindowManager *_instance;
         [preferencesWindowController.window close];
     }
     
+    // Workaround for SwiftUI Settings window
+    // https://github.com/tisfeng/Easydict/issues/362
+    for (NSWindow *window in [NSApp windows]) {
+        if ([window.identifier isEqualToString:@"com_apple_SwiftUI_Settings_window"]) {
+            [window close];
+        }
+    }
+    
     // get safe window position
     CGPoint safeLocation = [EZCoordinateUtils getFrameSafePoint:window.frame moveToPoint:point inScreen:self.screen];
     [window setFrameOrigin:safeLocation];
@@ -903,11 +911,6 @@ static EZWindowManager *_instance;
     
     self.floatingWindow.titleBar.pin = NO;
     [self.floatingWindow close];
-    
-    if (![EZPreferencesWindowController.shared isShowing]) {
-        // recover last app.
-        [self activeLastFrontmostApplication];
-    }
     
     if ([EZMainQueryWindow isAlive]) {
         [self.mainWindow orderBack:nil];
--- a/Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
+++ b/Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
@@ -408,18 +408,7 @@ static EZWindowManager *_instance;
         return;
     }
     
-    EZPreferencesWindowController *preferencesWindowController = [EZPreferencesWindowController shared];
-    if (preferencesWindowController.isShowing) {
-        [preferencesWindowController.window close];
-    }
-    
-    // Workaround for SwiftUI Settings window
-    // https://github.com/tisfeng/Easydict/issues/362
-    for (NSWindow *window in [NSApp windows]) {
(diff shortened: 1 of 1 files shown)


Jerry23011 (dismissed):

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
```diff
@@ -904,11 +912,6 @@ - (void)closeFloatingWindow {
     self.floatingWindow.titleBar.pin = NO;
     [self.floatingWindow close];
     
-    if (![EZPreferencesWindowController.shared isShowing]) {
-        // recover last app.
-        [self activeLastFrontmostApplication];
```
tisfeng, line 909: This is code is needed, we should active the last frontmost app when closing Easydict query window if the settings window is not currently displaying.

Does this code cause any problems?

tisfeng (changes requested):

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
```diff
@@ -413,6 +413,14 @@ - (void)showFloatingWindow:(EZBaseQueryWindow *)window atPoint:(CGPoint)point {
         [preferencesWindowController.window close];
     }
     
+    // Workaround for SwiftUI Settings window
+    // https://github.com/tisfeng/Easydict/issues/362
+    for (NSWindow *window in [NSApp windows]) {
+        if ([window.identifier isEqualToString:@"com_apple_SwiftUI_Settings_window"]) {
```
tisfeng, line 419: This code works, but can we find a better solution instead of hard coding?

tisfeng (changes requested):

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
phlpsong: With this code, it will not work to use shortcut to show the settings window.

You could try to type 'cmd' + ',' when query window active, and check if settings window shown.

phlpsong: This change seems not effect the user experience per my test.

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
phlpsong: I'm not sure. IMHO, there is no better ways to handle the SwiftUI window in AppKit code.

tisfeng: ok

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
tisfeng: > With this code, it will not work to use shortcut to show the settings window.
> 
> You could try to type 'cmd' + ',' when query window active, and check if settings window shown.

I fixed this issue in 792769e3

tisfeng: Activating the last frontmost app is mainly to restore the cursor position, making it convenient for users to continue using. 

Imagine a user is editing and modifying a word within a paragraph, such as forgetting how to spell "cursor", after invoking Easydict window to get the result, he wants to continue typing where he left off.

phlpsong: Great! works well from my side.

tisfeng (approved):

PR #411, opened 2024-02-18
pref: add help menu

AkaShark:


CanglongCl (dismissed):

Easydict/NewApp/View/MenuView/MainMenuCommand.swift
```diff
@@ -9,8 +9,21 @@
...
         // shortcut
         MainMenuShortcutCommand()
+
+        // Help
+        CommandGroup(replacing: CommandGroupPlacement.help, addition: {
+            Button("menu_feedback") {
+                guard let versionURL = URL(string: "\(EZGithubRepoEasydictURL)/issues") else {
+                    return
+                }
+                openURL(versionURL)
```
phlpsong, line 25:
```suggestion
                openURL(URL(string: "\(EZGithubRepoEasydictURL)/issues")!)
```

Personally, I think it's safe to force unwrap a hardcode url here.

AkaShark: But if the URL error, the App will open a blank page in web.

phlpsong: This hardcoded string should not fail to init in this case. Otherwise it will generate a fatal error.

AkaShark: yeah, I think you are right!

phlpsong (approved):

CanglongCl (approved):

Jerry23011 (approved):
LGTM. We could also add Easydict Help there later

AkaShark:
> LGTM. We could also add Easydict Help there later, it's just that Apple's current help book framework is very annoying to implement

okay, I will implement it in the next PR.

tisfeng (approved):

PR #416, opened 2024-02-21
Add settings to shortcuts

NeverAgain11:
#354


Easydict/Feature/ViewController/View/Titlebar/EZTitlebar.m
```diff
@@ -199,6 +219,42 @@ - (void)updatePinButtonImage {
...
+- (void)clickAutomaticallyRemoveCodeCommentSymbols {
+    Configuration.shared.automaticallyRemoveCodeCommentSymbols = !Configuration.shared.automaticallyRemoveCodeCommentSymbols;
+}
+
+- (void)clickAutomaticWordSegmentation {
+    Configuration.shared.automaticWordSegmentation = !Configuration.shared.automaticWordSegmentation;
+}
+
+- (void)goToSettings {
+    [EZPreferencesWindowController.shared show];
```
tisfeng, line 256: Why open the old settings when clicking `Go to Settings`?

image

Easydict/Feature/ViewController/View/Titlebar/EZTitlebar.m
```diff
@@ -64,7 +67,26 @@ - (void)setup {
...
+    
+}
+
+- (void)setupSettingButton {
+    EZOpenLinkButton *button = [[EZOpenLinkButton alloc] init];
+    NSImage *image = [[NSImage imageWithSystemSymbolName:@"switch.2" accessibilityDescription:nil]imageWithSymbolConfiguration:[NSImageSymbolConfiguration configurationWithScale:NSImageSymbolScaleLarge]];
+    
+    button.image = image;
+    self.settingButton = button;
+    button.clickBlock = nil;
```
tisfeng, line 82: This settings button, should also allow users to hide it, just like the Google link button.

Easydict/App/Localizable.xcstrings
```diff
@@ -485,6 +485,38 @@
         }
       }
     },
+    "automatic_word_segmentation" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Automatic word segmentation"
```
tisfeng, line 493:
```suggestion
            "value" : "English word auto-splitting"
```

Easydict/App/Localizable.xcstrings
```diff
@@ -485,6 +485,38 @@
...
+          }
+        }
+      }
+    },
+    "automatically_remove_code_comment_symbols" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Automatically remove code comment symbols."
```
tisfeng, line 509:
```suggestion
            "value" : "Automatically remove code comment symbols"
```

tisfeng:
The highlighted state of this settings button looks a bit strange compared to the Eudic button next to it, and after clicking on "Go to Settings", this highlighted state does not disappear.

image

Jerry23011 (commented):
Looks awesome, I just got two issues:
1. The colour pattern of the switch button is in default, so it is blacker than the pin icon on the right, can we customize it to match the colour of the pin icon?
2. Do we want to add an option in settings to allow users disable the button or keep it there as an option

AkaShark:
CleanShot 2024-02-22 at 14 07 15@2x
This settings button looks like bigger than other quick link button. May be you can use `resizeImage` make it better 
```Objective-C
   EZOpenLinkButton *button = [[EZOpenLinkButton alloc] init];
    NSImage *image = [NSImage imageWithSystemSymbolName:@"switch.2" accessibilityDescription:nil];
    image = [image resizeToSize:CGSizeMake(16, 16)];
    NSColor *tintColor = [NSColor mm_colorWithHexString:@"#C0C1C4"];
    image = [image imageWithTintColor:tintColor];
    button.image = image;
```
CleanShot 2024-02-22 at 14 34 49@2x

AkaShark (changes requested):
CleanShot 2024-02-22 at 14 07 15@2x
This settings button looks like bigger than other quick link button. May be you can use `resizeImage` make it better 
```Objective-C
   EZOpenLinkButton *button = [[EZOpenLinkButton alloc] init];
    NSImage *image = [NSImage imageWithSystemSymbolName:@"switch.2" accessibilityDescription:nil];
    image = [image resizeToSize:CGSizeMake(16, 16)];
    NSColor *tintColor = [NSColor mm_colorWithHexString:@"#C0C1C4"];
    image = [image imageWithTintColor:tintColor];
    button.image = image;
```
CleanShot 2024-02-22 at 14 34 49@2x

NeverAgain11:
Fixed problem with setting button.
@Jerry23011 @AkaShark @tisfeng

Easydict/Feature/ViewController/View/Titlebar/EZTitlebar.m
NeverAgain11: It doesn't seem to be able to open settings programmatically in MacOS 14.

tisfeng: I am not very familiar with SwiftUI Settings, and I'm not sure if there is a convenient solution here.

If there is no better solution, I have a simple and rude idea here, which is to open the settings by simulating the shortcut `Cmd + ,`, like this

```objc
[EZSystemUtility postKeyboardEvent:kCGEventFlagMaskCommand virtualKey:kVK_ANSI_Comma keyDown:true];
```

phlpsong: I made some changes in my branch(https://github.com/phlpsong/Easydict/tree/open_settings_appkit) to achieve this. I can create a PR later if this workaround is acceptable.

tisfeng: No, it's not good, open url `easydictd://settings` will query word "settings" in Easydict rather than open settings.

Also, using the URL Scheme method is not convenient for debugging because we are not sure if it will open the debug or release Easydict App.

phlpsong: URL schema does not works well after we register the handler for AppleEvent.  Currently, I am using notification to handle this. You could check the latest code.

tisfeng: Seems no bad, it works as expected although a bit complicated.

Please submit a PR, let's review the code carefully.

phlpsong: Yes, indeed.

PR created: https://github.com/NeverAgain11/Easydict/pull/1

NeverAgain11: PR was merged

Jerry23011 (dismissed):
Looks amazing, thank you

AkaShark (dismissed):

phlpsong (approved):

AkaShark (dismissed):

phlpsong (dismissed):

tisfeng (changes requested):
The color of the Settings button doesn't seem to match the color of the Google button next to it, what do you think about changing the TintColor to light blue? Or another suitable color.

image

image

Easydict/Feature/ViewController/View/Titlebar/EZTitlebar.m
```diff
@@ -199,6 +221,46 @@ - (void)updatePinButtonImage {
...
 
+- (void)showMenu {
+    NSMenu * menu = [[NSMenu alloc]initWithTitle:@"Menu"];
+    NSMenuItem * item1 = [[NSMenuItem alloc]initWithTitle:NSLocalizedString(@"automatically_remove_code_comment_symbols", nil) action:@selector(clickAutomaticallyRemoveCodeCommentSymbols) keyEquivalent:@""];
+    item1.target = self;
+    if (Configuration.shared.automaticallyRemoveCodeCommentSymbols) {
+        item1.state = NSControlStateValueOn;
+    }
+    
+    NSMenuItem * item2 = [[NSMenuItem alloc]initWithTitle:NSLocalizedString(@"automatic_word_segmentation", nil) action:@selector(clickAutomaticWordSegmentation) keyEquivalent:@""];
```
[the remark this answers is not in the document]
tisfeng: @NeverAgain11 You could copy and move these two options to the settings page, and alternatively, change the two settings options currently here to actions on the query text.

For example, change it to [remove code comment symbols], and if the user clicks on it, then execute `[queryText removeCommentBlockSymbols];`, essentially they are the same, except that one is executed automatically at translation time, and one is triggered manually by the user to process the query text.

tisfeng (changes requested):

Easydict/Feature/ViewController/Model/EZQueryModel.m
```diff
@@ -186,20 +186,22 @@ - (NSString *)handleInputText:(NSString *)inputText {
...
+                    queryText = [queryText splitCodeText];
+                }
             }
         }
     }
     
-    if (Configuration.shared.beta) {
+    if (Configuration.shared.automaticallyRemoveCodeCommentSymbols) {
         // Remove prefix [//,#,*,] and join texts.
         queryText = [queryText removeCommentBlockSymbols];
```
tisfeng, line 206: These options are actually all about processing the query text.

tisfeng (changes requested):

Jerry23011:
> The color of the Settings button doesn't seem to match the color of the Google button next to it, what do you think about changing the TintColor to light blue? Or another suitable color.

I would personally prefer to keep it grey. Since blue represents selected in the pin feature, it could be confusing to make the icon blue.

Also, for users who don't enable the other icons, the grey one looks more concise as it fits the accent colour of the app.
截屏2024-02-26 21 49 47

tisfeng:
Sounds reasonable, what do the rest of you think about this?

phlpsong:
I think grey is better.

AkaShark:
Agreed it!

Easydict/Feature/ViewController/View/Titlebar/EZTitlebar.m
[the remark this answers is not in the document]
NeverAgain11: I'm busy at work and may not have time to make these changes.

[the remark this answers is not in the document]
tisfeng: It's okay to do this on the weekends.

Usually these issues are not urgent, so if you are busy with work, you can put them off, just specify.

tisfeng:
Ok, let's keep it grey.

Easydict/Feature/ViewController/View/Titlebar/EZTitlebar.m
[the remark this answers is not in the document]
NeverAgain11: Should we change the text in the input field while triggering a manual action?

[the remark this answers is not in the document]
tisfeng: Of course，just like `inputText = [queryText removeCommentBlockSymbols]`

tisfeng (dismissed):

tisfeng:
Looks good.

Easydict/App/Localizable.xcstrings
```diff
@@ -2335,6 +2383,22 @@
         }
       }
     },
+    "remove_code_comment_symbols" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Remove code comment symbols「/*#」"
```
phlpsong, line 2391: Could we omit the `「/*#」` and `key_value —> key value`  part to make these two menu items look shorter?

tisfeng: > Automatically remove code comment symbols「/*#」

I think it's ok, since we have already displayed it completely in the settings.

phlpsong: Yes, another one is I am not sure if using `Remove comment symbols` is enough here？

tisfeng (approved):

tisfeng:
Please continue to review.

AkaShark (approved):

Issue #421, opened 2024-02-24
🐞 Bug Report: Validate Bing service always fails


tisfeng:
### Please confirm the following:

- [X] Be sure to check out [FAQ](https://github.com/tisfeng/Easydict/wiki/FAQ)
- [X] I have carefully read the [README](https://github.com/tisfeng/Easydict#readme)
- [X] I have searched through the [issues](https://github.com/tisfeng/Easydict/issues) page but couldn't find a solution.(Including issue that has been closed)
- [X] Easydict has been upgraded to the [latest version](https://github.com/tisfeng/Easydict/releases)
- [X] I understand and agree to the above, and understand that the project maintainer has limited energy, **issues that do not follow the rules may be ignored or closed directly**

### Problem description

Validate Bing service always fails

### Is the issue consistently reproducible?

Reproducible

### Steps to Reproduce

*

### Screenshot

<img width="814" alt="image" src="https://github.com/tisfeng/Easydict/assets/25194972/b04e7c93-39e9-4a9f-bec0-9656f5fdfa75">

### Expected result

*

### Device type and OS version

*

### Are you willing to submit a PR to fix this issue?

- [X] I'm willing to submit a PR

tisfeng:
[2.6.1](https://github.com/tisfeng/Easydict/releases/tag/2.6.1) 版本已修复该问题。

PR #422, opened 2024-02-24
fix: validate Bing service always fails

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/421

### DIFF
--- a/Easydict/NewApp/Configuration/Configuration+Defaults.swift
+++ b/Easydict/NewApp/Configuration/Configuration+Defaults.swift
@@ -55,7 +55,7 @@ extension Defaults.Keys {
     static let clearInput = Key<Bool>("EZConfiguration_kClearInputKey", default: true)
     static let keepPrevResultWhenEmpty = Key<Bool>("EZConfiguration_kKeepPrevResultKey", default: true)
     static let selectQueryTextWhenWindowActivate = Key<Bool>("EZConfiguration_kSelectQueryTextWhenWindowActivate", default: false)
-    static let enableBetaNewApp = Key<Bool>("EZConfiguration_kEnableBetaNewAppKey", default: false)
+    static let enableBetaNewApp = Key<Bool>("EZConfiguration_kEnableBetaNewAppKey", default: true)
 
     static let enableBetaFeature = Key<Bool>("EZBetaFeatureKey", default: false)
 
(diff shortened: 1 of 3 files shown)


tisfeng:
Since refactoring `startQuery:`, we don't need to call `prehandleQueryTextLanguage:` method manually anymore.

```objc
- (void)startQuery:(EZQueryModel *)queryModel
        completion:(void (^)(EZQueryResult *result, NSError *_Nullable error))completion {
    NSString *queryText = queryModel.queryText;
    EZLanguage from = queryModel.queryFromLanguage;
    EZLanguage to = queryModel.queryTargetLanguage;
    
    if ([self prehandleQueryTextLanguage:queryText
                                    from:from
                                      to:to
                              completion:completion]) {
        return;
    }
    
    [self translate:queryText from:from to:to completion:completion];
}
```

tisfeng:
I plan to release a small fix update tonight or tomorrow.

Easydict/NewApp/Configuration/Configuration+Defaults.swift
```diff
@@ -55,7 +55,7 @@ extension Defaults.Keys {
     static let clearInput = Key<Bool>("EZConfiguration_kClearInputKey", default: true)
     static let keepPrevResultWhenEmpty = Key<Bool>("EZConfiguration_kKeepPrevResultKey", default: true)
     static let selectQueryTextWhenWindowActivate = Key<Bool>("EZConfiguration_kSelectQueryTextWhenWindowActivate", default: false)
-    static let enableBetaNewApp = Key<Bool>("EZConfiguration_kEnableBetaNewAppKey", default: false)
+    static let enableBetaNewApp = Key<Bool>("EZConfiguration_kEnableBetaNewAppKey", default: true)
```
tisfeng, line 58: Since there seems to be no problem with the new SwiftUI Setting, we changed it to be enabled by default. 

In order to make it easier to use the new Swift/SwiftUI features, I plan to change the minimum supported system from macOS 11 to macOS 13 in the next version. 

After that, we no longer need to maintain the old Setting related code and move forward lightly.

phlpsong (approved):

tisfeng:
We need to speed up the Swift refactoring and turn it from an objc project into a Swift project.

Next, in order to better add new features later, I plan to rewrite the code related to the query service (excluding the services themselves), the previous code was a bit ugly, when we refactor it later, we should try our best to optimize it, making it easier to expand.

image

AkaShark (approved):

PR #424, opened 2024-02-24
fix: service toggle issue after switch focus

phlpsong:
close https://github.com/tisfeng/Easydict/issues/423


tisfeng:
The code works, but it will cause this bug again https://github.com/tisfeng/Easydict/pull/326#issuecomment-1893017637

Easydict/NewApp/View/SettingView/Tabs/ServiceTab.swift
```diff
@@ -121,28 +121,29 @@ private struct ServiceItems: View {
...
 private struct ServiceItemView: View {
     let service: QueryService
 
     @EnvironmentObject private var viewModel: ServiceTabViewModel
 
-    private var enabled: Binding<Bool> {
-        .init {
-            service.enabled
-        } set: { newValue in
-            guard service.enabled != newValue else { return }
```
tisfeng, line 134: This code is needed, we shouldn't post update notification if enable has no new value.

tisfeng (changes requested):

Easydict/NewApp/View/SettingView/Tabs/ServiceTab.swift
phlpsong: The logic moved inside of `onChangeValue` modifier.

phlpsong:
> The code works, but it will cause this bug again [#326 (comment)](https://github.com/tisfeng/Easydict/pull/326#issuecomment-1893017637)

fixed.

tisfeng:
ok, looks good.

tisfeng (approved):

Jerry23011 (approved):
LGTM. Also fixed a bug where the toggle state is not updating after switching windows.

AkaShark (approved):

PR #427, opened 2024-02-25
style: use full width colon in zh_CN

phlpsong:


AkaShark (dismissed):

Easydict/App/Localizable.xcstrings
```diff
@@ -4048,7 +4048,7 @@
         "zh-Hans" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "标签:"
+            "value" : "标签："
```
tisfeng, line 4051: It's ok to change the colon in the settings page to full-width, but for the query display result part, the full-width `：` is much wider than the half-width `:`, which will lead to some difference in the display effect in Chinese and English environments, and it is not suitable for modifying the spacing of the controls in the code, so I suggest not to modify this part.

image

tisfeng (changes requested):

Easydict/App/Localizable.xcstrings
phlpsong: I did some research on several translation services and found that most of them do not use a colon before phonetic symbols but use a full-width colon in other parts.

截屏2024-02-27 12 45 48

I wonder if we could optimize here like most other services like omitting the colon before phonetic and removing the label before the description part and tag part. I think these items are quite clear for users after removing labels.

tisfeng: OK, we can remove the colon in front of phonetic, as well as the tag label and explanation label.

tisfeng: Remember to remove the phonetic colon in the English environment as well.

Easydict/App/Localizable.xcstrings
```diff
@@ -1082,7 +1082,7 @@
         "zh-Hans" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "词源:"
+            "value" : "词源："
```
tisfeng, line 1085: This doesn't need to be modified, in fact, it's not being used at the moment. 

When rewriting with Swift later on, this part of the code can be removed.

tisfeng (changes requested):

Easydict/App/Localizable.xcstrings
phlpsong: sure, will update later.

Easydict/App/Localizable.xcstrings
phlpsong: Maybe keep it for now?  We could delete it when once the label has been removed.

tisfeng: Yes, keep it.

tisfeng (approved):

AkaShark (dismissed):

AkaShark (approved):

tisfeng (approved):

PR #428, opened 2024-02-26
Introduce `SwiftLint` and `.swiftformat`

CanglongCl:
Most of changes are made by `swiftformat` which do not influence functions of the code.


.swiftformat
```diff
@@ -1,4 +1,108 @@
-# https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md
+# SwiftFormat config compliant with Google Swift Guideline
+# https://google.github.io/swift/#control-flow-statements
 
-# file options
---exclude Pods
\ No newline at end of file
+# Specify version used in a project
+
+--swiftversion 5.7
```
tisfeng, line 6: Can we use swift version 5.9 ?

tisfeng: I have added a `.swift-version` file  to specify swift version in 41bcc869 , if use `--swiftversion`, can we remove that file?

CanglongCl: This file also contains other rules which is consistent with swiftlint configuration. You may edit rules in this file but we'd better not to remove it.

tisfeng: ok

tisfeng (dismissed):
Looks good.

phlpsong (changes requested):

.swiftlint.yml
```diff
@@ -0,0 +1,71 @@
...
+  warning: 900
+  error: 1600
+type_name:
+  min_length: 3
+  max_length:
+    warning: 50
+    error: 50
+identifier_name:
+  min_length: 3
+  excluded: # excluded via string array
```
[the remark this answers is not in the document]
CanglongCl: No. Functions like 
```swift
override func translate(
    _ text: [String],
    from: Language,
    to: Language,
    completion: @escaping (EZQueryResult, Error?) -> ()
)
```
should be written as like 
```swift
override func translate(
    _ text: [String],
    from sourceLanguage: Language,
    to destinationLanguage: Language,
    completion: @escaping (EZQueryResult, Error?) -> ()
)
```
which is a more clear identifier.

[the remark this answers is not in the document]
phlpsong: ok

Easydict/Feature/Service/Ali/AliService.swift
```diff
@@ -37,13 +25,8 @@ class AliService: QueryService {
...
     }
 
-    override func hasPrivateAPIKey() -> Bool {
-        let id = Defaults[.aliAccessKeyId] ?? ""
-        let secret = Defaults[.aliAccessKeySecret] ?? ""
-        return !id.isEmpty && !secret.isEmpty
-    }
-
     override public func supportLanguagesDictionary() -> MMOrderedDictionary<AnyObject, AnyObject> {
+        // swiftlint:disable:next todo
```
phlpsong, line 29:
```suggestion
```

CanglongCl: @tisfeng How do you think of the `todo` rule? Maybe leaving it is a better choice?

phlpsong: I believe it will not trigger a warning after removing this line.

phlpsong (approved):

tisfeng (approved):

PR #433, opened 2024-02-27
add a custom OpenAI service

phlpsong:
close https://github.com/tisfeng/Easydict/issues/387


tisfeng (changes requested):
This icon looks bad in dark mode.

image

Easydict/NewApp/Utility/Extensions/QueryService+ConfigurableService/OpenAIService+ConfigurableService.swift
```diff
@@ -21,12 +21,6 @@ extension EZOpenAIService: ConfigurableService {
                 key: .openAIAPIKey,
                 placeholder: "service.configuration.openai.api_key.placeholder"
             )
-            // endpoint
-            ServiceConfigurationInputCell(
-                textFieldTitleKey: "service.configuration.openai.endpoint.title",
-                key: .openAIEndPoint,
-                placeholder: "service.configuration.openai.endpoint.placeholder"
-            )
```
tisfeng, line 29: Although a new one-api service has been added this time, I don't think we should remove the endpoint from the previous OpenAI, which will cause the existing configuration service to not work.

image

tisfeng (changes requested):
This error message is strange, I just haven't configured the one-api service.

image

Easydict/App/Localizable.xcstrings
```diff
@@ -1970,6 +1970,23 @@
         }
       }
     },
+    "one_api" : {
+      "comment" : "The name of One API Translate",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "One API Translate"
```
tisfeng, line 1979:
```suggestion
            "value" : "Custom OpenAI Translate"
```
How about renaming it to `Custom OpenAI Translate`?

`One API Translate` is not clear, I want user to know it's a general OpenAI service.

Easydict/App/Localizable.xcstrings
```diff
@@ -1970,6 +1970,23 @@
...
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "One API 翻译"
```
tisfeng, line 1985:
```suggestion
            "value" : "自定义 OpenAI 翻译"
```

Easydict/App/Localizable.xcstrings
phlpsong: ok

Easydict/NewApp/Utility/Extensions/QueryService+ConfigurableService/OpenAIService+ConfigurableService.swift
phlpsong: But keeping this will make there minor differences between the new service and the OpenAI service.

Jerry23011 (changes requested):
截屏2024-03-04 20 39 03
After updating the localization, it seems that the switch is being relocated.

phlpsong:
> 截屏2024-03-04 20 39 03 After updating the localization, it seems that the switch is being relocated.

It's strange which en string is longer than ch but its behavior is fine.  I will add a minWidth constraint to fix this.

phlpsong:
> This icon looks bad in dark mode.
> 
> image

I have made a simple one to replace it, let me know what you think.

Easydict/Feature/Service/OneAPI/OneAPIService.swift
```diff
@@ -0,0 +1,118 @@
...
+//  Created by phlpsong on 2024/2/16.
+//  Copyright © 2024 izual. All rights reserved.
+//
+
+import Alamofire
+import CryptoKit
+import Defaults
+import Foundation
+
+@objc(EZOneAPIService)
```
tisfeng, line 14: Please rename all `OneAPIService` to `CustomOpenAIService`

Easydict/App/Localizable.xcstrings
```diff
@@ -2673,82 +2690,98 @@
...
       }
     },
-    "service.configuration.openai.api_key.title" : {
+    "service.configuration.oneapi.model.placeholder" : {
       "localizations" : {
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "API Key"
+            "value" : "gpt-3.5-turbo/gemini-pro"
```
tisfeng, line 2714:
```suggestion
            "value" : "gpt-3.5-turbo"
```

Easydict/App/Localizable.xcstrings
```diff
@@ -2673,82 +2690,98 @@
...
           }
         },
         "zh-Hans" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "API Key"
+            "value" : "gpt-3.5-turbo/gemini-pro"
```
tisfeng, line 2720:
```suggestion
            "value" : "gpt-3.5-turbo"
```

tisfeng:
Looks bad in English mode.

If it's not appropriate to increase the width here, we could shorten the title spelling, `Custom OpenAl Translate` --> `Custom OpenAl`, which I think users should be able to understand.

image

Easydict/Feature/Service/OneAPI/OneAPIService.swift
phlpsong: Already changed in prev commits.

phlpsong:
> Looks bad in English mode.
> 
> If it's not appropriate to increase the width here, we could shorten the title spelling, `Custom OpenAl Translate` --> `Custom OpenAl`, which I think users should be able to understand.
> 
> image

I think it's already fixed in prev commits. Could you build using the latest commit and take a look?

Easydict/NewApp/Utility/Extensions/QueryService+ConfigurableService/OpenAIService+ConfigurableService.swift
phlpsong: ok, already reverted the changes of endPoint part.

Easydict.xcodeproj/project.pbxproj
```diff
@@ -3469,7 +3469,7 @@
 				COPY_PHASE_STRIP = NO;
 				CURRENT_PROJECT_VERSION = 33;
 				DEAD_CODE_STRIPPING = YES;
-				DEVELOPMENT_TEAM = "$(DEVELOPMENT_TEAM)";
+				DEVELOPMENT_TEAM = XXM539CLCH;
```
tisfeng, line 3472: Why change this?

tisfeng (changes requested):

Easydict.xcodeproj/project.pbxproj
tisfeng: image

phlpsong: submit by mistake, let me revert this🤡

phlpsong:
> It's working well enough so far 😌
> 
> 
> 
> For custom OpenAI services, how about we show the model currently being used to the right of the header?
> 
> 
> 
> image
> 
> 

Good idea! Does OpenAI service need to apply this?

tisfeng:
I'm not sure how the display effect of OpenAI will be, because the names of the two preset models are a bit long, but we can display it first and see the effect.

image

phlpsong:
> I'm not sure how the display effect of OpenAI will be, because the names of the two preset models are a bit long, but we can display it first and see the effect.
> 
> image

Screenshot 2024-03-06 at 19 28 42

`Custom OpenAI Translate` is too long in the mini & fix window which has the overlap issue, but it works well in Chinese.

Easydict/Feature/ViewController/View/ResultView/EZResultView.m
```diff
@@ -14,12 +14,14 @@
...
 #import "NSImage+EZSymbolmage.h"
 #import "EZWindowManager.h"
+#import "EZOpenAILikeService.h"
 
 @interface EZResultView ()
 
 @property (nonatomic, strong) NSView *topBarView;
 @property (nonatomic, strong) NSImageView *serviceIcon;
 @property (nonatomic, strong) NSTextField *serviceNameLabel;
+@property (nonatomic, strong) EZButton *serviceModelButton;
```
tisfeng, line 24: I saw that you replaced `NSTextField` with `EZButton`, is there any issue with using a label?

Easydict/Feature/ViewController/View/ResultView/EZResultView.m
```diff
@@ -361,6 +396,21 @@ - (void)updateArrowButton {
     }];
 }
 
+- (BOOL)isOpenAIService:(EZQueryService *)service {
+    EZServiceType type = service.serviceType;
+    return type == EZServiceTypeCustomOpenAI || type == EZServiceTypeOpenAI;
```
tisfeng, line 401: Would it be better to judge whether it is an OpenAI type of service? Later we need to support custom OpenAI services https://github.com/tisfeng/Easydict/issues/439#issuecomment-1977974655 .

```objc
[service isKindOfClass:EZOpenAILikeService.class]
```

Easydict/Feature/ViewController/View/ResultView/EZResultView.m
phlpsong: No issues with a label, but I changed the appearance to have a background and cornerRadius, the button seems easy to customize.

tisfeng (changes requested):
`Custom OpenAl Translate` looks too long in English, how about adding a `Service Name` option to enable user to customize service name?

Moreover, I also found that even for the same model, it might come from different service providers, adding a custom service name will help the users to differentiate this.

image

Easydict/Feature/ViewController/View/ResultView/EZResultView.m
tisfeng: It's ok to use button here.

I just thought that we can use this button to switch models easily. When the button is clicked, an `NSMenu` will pop up, which is very useful for quickly switching models.

image

tisfeng:
We can change Model to Models, users can manually enter multiple models separated by `,`, we internally save them in an array, the first one will be used as the selected model by default, and users can conveniently switch between models in the query window.

This is just my initial idea, there may be some things that are not well considered and need to be improved.

image

image

Easydict/Feature/ViewController/View/ResultView/EZResultView.m
phlpsong: It will be easy to switch the model for openAI service since it has pre defined models. But for custom openai service, there is no predefined model list. I will check if have any good ways to handle this.

phlpsong:
I think using the comma to separate models is a bit complicated for users. I will check this later if have any other good ways to handle it.

tisfeng:
In my opinion, letting the user manually input the custom model is already a relatively simple and general enough solution.

I know some services support API querying model lists, but the custom OpenAI service that we designed is very freestyle and shouldn't assume that it will support API querying model lists.

On the other hand, even if it did support API querying model lists, it's not necessarily suitable to directly use it in Easydict to display the `NSMenu`, because the number of models that some services support is very large, for example OpenAI or [Together AI](https://docs.together.ai/docs/openai-api-compatibility),  they return more than a dozen models., and not every model is suitable for translation.

If you have a better idea, let me know.

phlpsong:
> In my opinion, letting the user manually input the custom model is already a relatively simple and general enough solution.
> 
> I know some services support API querying model lists, but the custom OpenAI service that we designed is very freestyle and shouldn't assume that it will support API querying model lists.
> 
> On the other hand, even if it did support API querying model lists, it's not necessarily suitable to directly use it in Easydict to display the `NSMenu`, because the number of models that some services support is very large, for example OpenAI or [Together AI](https://docs.together.ai/docs/openai-api-compatibility), they return more than a dozen models., and not every model is suitable for translation.
> 
> If you have a better idea, let me know.

Yes, I agree with you, there's no better way to do this. 

I have implemented this logic in the latest commit, you could kindly take a look and let me know if you have any comments.

Easydict/Feature/Service/CustomOpenAI/CustomOpenAIService.swift
```diff
@@ -0,0 +1,118 @@
...
+        NSLocalizedString("custom_openai", comment: "The name of Custom OpenAI Translate")
+    }
+
+    override public func link() -> String? {
+        "https://chat.openai.com"
+    }
+
+    // MARK: Internal
+
+    static let defaultModels = ["gpt-3.5-turbo-0125", "gpt-4-0125-preview"]
```
tisfeng, line 41: No, we don't need to set default models here, just show placeholder enough.

Also, it looks like there's a bug here,  if you delete all the models, it gets set to the default models again.

image

tisfeng (changes requested):

tisfeng:
> For models quick switch function in the query window, I could make this enhancement in a separate PR later. @tisfeng

Ok, this PR has enough content, let's put the rest of the features in another PR.

Easydict/Feature/Service/CustomOpenAI/CustomOpenAIService.swift
phlpsong: The behavior is by design since I think it need to assign a default model here. 

I will update this part later.

tisfeng (changes requested):
After the custom service name is changed, the name displayed in the list is not updated.

image

Easydict/NewApp/Utility/Extensions/QueryService+ConfigurableService/CustomOpenAIService+ConfigurableService.swift
```diff
@@ -0,0 +1,155 @@
...
+        }
+        validModels = availableModels.components(separatedBy: ",").filter { !$0.isEmpty }
+        if validModels.count == 1 || !validModels.contains(model) {
+            model = validModels[0]
+        }
+    }
+
+    private func serviceConfigChanged() {
+        let userInfo: [String: Any] = [EZWindowTypeKey: service.windowType.rawValue]
+        let notification = Notification(name: .serviceHasUpdated, object: nil, userInfo: userInfo)
```
tisfeng, line 152:
```suggestion
        let notification = Notification(name: .serviceHasUpdated, object: nil, userInfo: nil)
```

All windows currently share these services, so when you modify the service configuration, you need to update all window service information.

When userInfo is nil, it means to send update notification to all windows.

phlpsong: I think it's ok to pass the window type based on my test. The main window is updated while I edit the service info in the mini window segment.

Easydict/NewApp/Utility/Extensions/QueryService+ConfigurableService/CustomOpenAIService+ConfigurableService.swift
```diff
@@ -0,0 +1,154 @@
...
+import SwiftUI
+
+// MARK: - CustomOpenAIService + ConfigurableService
+
+@available(macOS 13.0, *)
+extension CustomOpenAIService: ConfigurableService {
+    func configurationListItems() -> some View {
+        ServiceConfigurationSecretSectionView(
+            service: self,
+            observeKeys: [.customOpenAIAPIKey, .customOpenAIEndPoint]
```
tisfeng, line 21: This should require input of the model before validation.

image

tisfeng (changes requested):

tisfeng:
I just removed the default API key in custom OpenAI service e763cc5f, I plan to add a separate built-in AI translation service later.

Easydict/NewApp/Utility/Extensions/QueryService+ConfigurableService/CustomOpenAIService+ConfigurableService.swift
phlpsong: fixed.

phlpsong:
> I just removed the default API key in custom OpenAI service [e763cc5](https://github.com/tisfeng/Easydict/commit/e763cc5f3672eaa113dee35e629a69fcf2e25525), I plan to add a separate built-in AI translation service later.

ok, I will take a look at the name update issue this day later.

tisfeng:
Every time the settings service page is entered, `serviceConfigChanged()` is triggered multiple times, posting update service notifications, which can cause the loss of query content.


https://github.com/tisfeng/Easydict/assets/25194972/2219afb0-c07b-4114-afca-5db4dacb096c

phlpsong:
> Every time the settings service page is entered, `serviceConfigChanged()` is triggered multiple times, posting update service notifications, which can cause the loss of query content.
> 
>  iShot_2024-03-10_13.49.19.mp4

Yes, I found this issue too, checking it now.

phlpsong:
The notification trigger issue still exists, need more time than expected to investigate it.

tisfeng:
ok, you can check this problem carefully.

Easydict/Feature/ViewController/Window/BaseQueryWindow/EZBaseQueryViewController.m
```diff
@@ -245,11 +245,31 @@ - (void)dealloc {
...
         [self updateServices];
     }
 }
 
+- (void)updateService:(NSString *)serviceType {
+    NSMutableArray<EZQueryService *> *newServices = [[NSMutableArray alloc] init];
+    for (EZQueryService *service in self.services) {
+        if ([service serviceType] == serviceType) {
+            EZQueryService *item = [EZLocalStorage.shared service:serviceType];
+            [newServices addObject:item];
```
tisfeng, line 263: When updating serve with this method, the new initialized serve doesn’t set `queryModel` and `windowType`, which will cause serve to fail to make a new query.

You need to improve `- (void)updateService:(NSString *)serviceType` by referring to `- (void)setupServices`.

Besides, it’s recommended that you refactor `- (void)setupServices` with the new `- (void)updateService:(NSString *)serviceType` method to decrease repeated code.

image

tisfeng (changes requested):

Easydict/Feature/ViewController/Window/BaseQueryWindow/EZBaseQueryViewController.m
phlpsong: Sure, changes have been updated, this issue should be fixed.

Easydict/NewApp/Utility/Extensions/QueryService+ConfigurableService/CustomOpenAIService+ConfigurableService.swift
```diff
@@ -149,7 +145,11 @@ private class CustomOpenAIViewModel: ObservableObject {
     }
 
     private func serviceConfigChanged() {
-        let notification = Notification(name: .serviceHasUpdated, object: nil, userInfo: nil)
+        let userInfo: [String: Any] = [
+            EZWindowTypeKey: service.windowType.rawValue,
```
tisfeng, line 167: The problem is back again https://github.com/tisfeng/Easydict/pull/433#discussion_r1518580949

Easydict/Feature/ViewController/Storage/EZLocalStorage.m
```diff
@@ -141,6 +141,20 @@ - (void)setAllServiceTypes:(NSArray<EZServiceType> *)allServiceTypes windowType:
...
+    BOOL enabledQuery = YES;
+    if (serviceInfo) {
+        enabled = serviceInfo.enabled;
+        enabledQuery = serviceInfo.enabledQuery;
+    }
+    service.enabled = enabled;
+    service.enabledQuery = enabledQuery;
+    return service;
+}
+
```
tisfeng, line 157: If you want to add this kind of repetitive code, it is recommended to refactor it as well.

```objc
- (NSArray<EZQueryService *> *)allServices:(EZWindowType)windowType {
    NSArray *allServices = [EZServiceTypes.shared servicesFromTypes:[self allServiceTypes:windowType]];
    for (EZQueryService *service in allServices) {
        EZServiceInfo *serviceInfo = [self serviceInfoWithType:service.serviceType windowType:windowType];
        BOOL enabled = YES;
        BOOL enabledQuery = YES;
        if (serviceInfo) {
            enabled = serviceInfo.enabled;
            enabledQuery = serviceInfo.enabledQuery;
        }
        service.enabled = enabled;
        service.enabledQuery = enabledQuery;
    }
    return allServices;
}

- (EZQueryService *)service:(EZServiceType)serviceType windowType:(EZWindowType)windowType {
    EZQueryService *service = [EZServiceTypes.shared serviceWithType:serviceType];
    EZServiceInfo *serviceInfo = [self serviceInfoWithType:service.serviceType windowType:windowType];
    BOOL enabled = YES;
    BOOL enabledQuery = YES;
    if (serviceInfo) {
        enabled = serviceInfo.enabled;
        enabledQuery = serviceInfo.enabledQuery;
    }
    service.enabled = enabled;
    service.enabledQuery = enabledQuery;
    return service;
}
```

Easydict/NewApp/Utility/Extensions/QueryService+ConfigurableService/CustomOpenAIService+ConfigurableService.swift
phlpsong: Yes, just realized it will have issues when different query windows shown. I will update this part.

Easydict/Feature/ViewController/Window/BaseQueryWindow/EZBaseQueryViewController.m
```diff
@@ -245,13 +244,31 @@ - (void)dealloc {
...
+        [self updateSingleService:serviceType windowType:type];
+        return;
+    }
     if (type == self.windowType || !userInfo) {
-        [self updateServices];
+        [self updateServices:[EZLocalStorage.shared allServices:self.windowType]];
+    }
+}
+
+- (void)updateSingleService:(NSString *)serviceType windowType:(EZWindowType)windowType {
```
tisfeng, line 257: This method looks good, but the actual effect may not be good, because the three query windows are currently independent, and they all share these services, so all window services usually need to be updated.

Easydict/Feature/ViewController/Window/BaseQueryWindow/EZBaseQueryViewController.m
```diff
@@ -245,13 +244,31 @@ - (void)dealloc {
...
+    NSMutableArray<EZQueryService *> *newServices = [[NSMutableArray alloc] init];
+    for (EZQueryService *service in self.services) {
+        if ([service serviceType] == serviceType) {
+            EZQueryService *updatedService = [EZLocalStorage.shared service:serviceType windowType:windowType];
+            [newServices addObject:updatedService];
+        } else {
+            [newServices addObject:service];
+        }
     }
+    [self updateServices:newServices];
```
tisfeng, line 267: In addition, this method looks like it can update one service alone, but in fact it ends up updating all services.

phlpsong: Yes, the function name maybe confusing, this actually did update one single service in current service list and then pass the new service array to `updateServices `

Easydict/Feature/ViewController/Window/BaseQueryWindow/EZBaseQueryViewController.m
phlpsong: use `self.windowType` instead.

tisfeng:
Ok, it looks like there is no problem so far.

This PR code has a lot of code and has been modified a lot, so it is not easy to review.

Next time, we should cut it into smaller PRs to make it easier for everyone to review.

tisfeng (approved):

phlpsong:
> Ok, it looks like there is no problem so far.
> 
> 
> 
> This PR code has a lot of code and has been modified a lot, so it is not easy to review.
> 
> 
> 
> Next time, we should cut it into smaller PRs to make it easier for everyone to review.

Agree, it's a huge work both for author and reviewer.

tisfeng (approved):

tisfeng:
Please review this PR.

CanglongCl (approved):

tisfeng:
Great job!

tisfeng:
Next, I hope to be able to click on the model button to quickly switch models https://github.com/tisfeng/Easydict/issues/456

phlpsong:
sure, I will do it later.

PR #440, opened 2024-03-04
Language preference

choykarl:
#348


tisfeng (changes requested):
It works, but it didn't change the display language of the query window.

image

tisfeng:
This is a huge work, it does work indeed and the effect is very good.

I originally thought that this feature only required switching languages within the application, then reloading the language bundle to achieve it, or restarting app in a programmed way 🥹

Easydict/App/Localizable.xcstrings
```diff
@@ -1631,6 +1631,23 @@
...
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Language"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "语言"
```
tisfeng, line 1646:
```suggestion
            "value" : "显示语言"
```

tisfeng (changes requested):
Something wrong?

image

tisfeng (changes requested):

PR #444, opened 2024-03-06
Migrate Automatically Check for Updates toggle from About tab to General Tab

Jerry23011:
Please tell me if I CVed the code right :)

### DIFF
--- a/Easydict/App/Localizable.xcstrings
+++ b/Easydict/App/Localizable.xcstrings
@@ -11,16 +11,6 @@
         }
       }
     },
-    "(lastest_version %@)" : {
-      "localizations" : {
-        "zh-Hans" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "（最新版本 %@）"
-          }
-        }
-      }
-    },
     "about" : {
       "comment" : "about",
       "localizations" : {
@@ -321,6 +311,12 @@
     },
     "auto_check_update " : {
       "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Automatically check for app updates"
+          }
+        },
         "zh-Hans" : {
           "stringUnit" : {
             "state" : "translated",
@@ -658,6 +654,12 @@
     },
     "check_for_updates" : {
       "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Check for updates"
+          }
+        },
         "zh-Hans" : {
           "stringUnit" : {
             "state" : "translated",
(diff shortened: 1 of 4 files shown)


AkaShark (approved):

tisfeng (changes requested):
I think it would be better to place the AutoCheckUpdate option on the General tab, because it is a common feature for most apps.

The advanced tab should place some less common features,, or features in Easydict that are not well understood by the average user.

image

Easydict/App/Localizable.xcstrings
```diff
@@ -309,6 +309,22 @@
...
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Automatically check for application updates (Latest version %@)"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "needs_review",
+            "value" : "自动检查应用程序更新（最新版本 %@）"
```
tisfeng, line 323: The `Latest version` could keep in About tab, we can add a `Check update` button in this page, like 1Password.

This makes it convenient for users to manually check for updates on the About tab, and also provides an alternative way  besides the menu icon.

image

image

tisfeng (changes requested):

Easydict/App/Localizable.xcstrings
Jerry23011: For the check for updates button, I think [CodeEdit](https://github.com/CodeEditApp/CodeEdit/blob/main/CodeEdit/Features/Settings/SettingsView.swift) has a pretty good approach to put it in general
截屏2024-03-06 13 34 58

tisfeng: Ok, it looks great, let's follow it.

Jerry23011: I've updated the code to fit the style of CodeEdit
截屏2024-03-09 00 15 55
截屏2024-03-09 00 17 18

Jerry23011: I think we can perhaps rewrite the About view later to have a horizontal layout. It doesn't look very modern at the moment

tisfeng: Yes, I agree with you.

Easydict/App/Localizable.xcstrings
```diff
@@ -309,6 +309,22 @@
...
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Automatically check for app updates"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "自动检查应用程序更新"
```
tisfeng, line 323: `auto_check_update` already exists.

image

Jerry23011: These two cannot be deleted as they are both auto-generated. I check the code and found the previous one to be for the objc About view. I will push a commit to unify their translations.
截屏2024-03-09 14 25 20

phlpsong (approved):

Easydict/App/Localizable.xcstrings
tisfeng: ok

tisfeng (approved):
Looks good.

PR #450, opened 2024-03-10
fix: multi screen window show

AkaShark:
closed https://github.com/tisfeng/Easydict/issues/442

~~Add a picker in the general page to display the window as auto or fixed when translation finishes.~~


phlpsong (dismissed):
LGTM

Easydict/Feature/Utility/EZCoordinateUtils/EZCoordinateUtils.m
```diff
@@ -30,7 +31,11 @@ + (CGPoint)getSafeLocation:(CGRect)frame inScreen:(NSScreen *)screen {
 /// Make sure frame show in screen visible frame, return left-bottom postion frame.
 + (CGRect)getSafeAreaFrame:(CGRect)frame inScreen:(nullable NSScreen *)screen {
     if (!screen) {
-        screen = [self screenOfMousePosition];
+        if (Configuration.shared.showWindowMultiScreen == EZShowWindowMultiScreenAuto) {
+            screen = [self screenOfMousePosition];
+        } else {
+            screen = NSScreen.mainScreen; // EZShowWindowMultiScreenFixed fixed in main screen
```
tisfeng, line 37: No, it should not use mainScreen here, it should use the screen when selecting words.

tisfeng (changes requested):
Normally, the position of the word is basically the same as the mouse position, but if the user moves the mouse position quickly during the query, or even across multiple screens, the problem will be complicated, which I have not considered before.

I thought about it and felt that there is no need to add an additional option here. We should use the screen where the selected word is located, rather than the current mouse position.

Usually, the position of the word selection is basically consistent with the mouse position. However, if users quickly move the mouse during the query process, even across multiple screens, it will make this issue very complex, which I hadn't considered before.

After some thought, I feel that there is no need to add an additional option here. We should use the screen where the word selection is made instead of the current mouse position.

Adding an extra setting option may confuse users. Using the location of word selection aligns better with user intuition.

AkaShark:
okay I will fix it later

AkaShark:
I remove the picker from the settings page.  When the user starts the query I record the mouse-located screen,  and remove this record when the floating window closes.

AkaShark:
> I remove the picker from the settings page. When the user starts the query action I record the mouse-located screen and remove this record when the floating window closes.

tisfeng (changes requested):
I tried it, but the current code does not work https://github.com/tisfeng/Easydict/issues/442#issuecomment-1987531812 :

Get the word on screen A, then move the mouse to screen B, and then use the shortcut keys to translate the word. The query window should be displayed at the word pick-up position on screen A.

tisfeng:
This problem can be a bit complex, involving multiple screens of records and the handling of the location of the display query window.

If you're not good at solving it, I can take over later.

AkaShark:
Oh, I now understand your issue. I didn't consider issue #442  earlier and only focused on the mouse move duration during query translation. To address this, I will record the screen when starting a query like this
```
- (void)startQueryText:(NSString *)text actionType:(EZActionType)actionType {
    // record current screen when start query
    NSScreen *current = [EZCoordinateUtils screenForPoint: [NSEvent mouseLocation]];
    EZCoordinateUtils.startQueryScreen = current;
```
Let me try it first. I'll ask for your help if I find it difficult for me.

AkaShark:
@tisfeng  My idea is to record the screen when a word is selected for translation and display a query window on this screen. For other modes of translation, the screen will still be recorded when the query is initiated. What do you think about it?

tisfeng:
I'm not so sure, you can try it and then we'll test it.

AkaShark:
> I'm not so sure, you can try it and then we'll test it.

okk

tisfeng (changes requested):
It still doesn't work.

Although the latest code can record screen A https://github.com/tisfeng/Easydict/pull/450#pullrequestreview-1931433425 , the position of the query window displayed when using the shortcut key to stroke the words to translate is inaccurate, and there is a problem that the query window flickers between screen A and screen B. 

It's not easy for me to record a video to show you this problem,  I suggest you test it yourself using multiple screens, its very easy to reproduce.

tisfeng:
@AkaShark  When you have time, please continue pushing this PR.

AkaShark:
> @AkaShark When you have time, please continue pushing this PR.

okay, I will try to fix it in the last few days

AkaShark:
I'm sorry I've been a little busy lately. I'll finish this issue as soon as possible.

tisfeng:
ok, you can schedule as you see fit.

Issue #456, opened 2024-03-16
dev: Click the model button to quickly switch OpenAI models


tisfeng:
I just thought that we can use this button to switch models easily. When the button is clicked, an `NSMenu` will pop up, which is very useful for quickly switching models.

https://github.com/tisfeng/Easydict/pull/433#discussion_r1515461839

<img width="439" alt="image" src="https://github.com/tisfeng/Easydict/assets/25194972/50bf73e8-eaee-4b5c-ab50-db5b27e545da">

PR #457, opened 2024-03-16
Switch model in query window

phlpsong:
close https://github.com/tisfeng/Easydict/issues/456


Easydict/Feature/ViewController/View/ResultView/EZResultView.m
```diff
@@ -409,6 +415,30 @@ - (void)updateServiceModelLabel {
...
+        [menu addItem:item];
+    }
+    [menu popUpMenuPositioningItem:nil atLocation:[NSEvent mouseLocation] inView:nil];
+}
+
+- (void)modelDidSelected:(NSMenuItem *)sender {
+    EZOpenAILikeService *service = (EZOpenAILikeService *)self.result.service;
+    service.model = sender.title;
+    self.serviceModelButton.title = service.model;
+    [self postServiceUpdatedNotification:service.serviceType];
```
tisfeng, line 433:
```suggestion
     if (![service.model isEqualToString:sender.title]) {
        service.model = sender.title;
        self.serviceModelButton.title = service.model;
        [self postServiceUpdatedNotification:service.serviceType];
    }
```

tisfeng: If the selected model doesn't change, we don' have to update.

Easydict/Feature/Service/OpenAI/EZOpenAIService.m
```diff
@@ -62,6 +62,14 @@ - (NSString *)model {
     return model;
 }
 
+- (void)setModel:(NSString *)model {
+    [[NSUserDefaults standardUserDefaults] setValue:model forKey:EZOpenAIModelKey];
+}
+
+- (NSArray<NSString *> *)availableModels {
+    return @[@"gpt-3.5-turbo-0125", @"gpt-4-0125-preview"];
```
tisfeng, line 70: Can we optimize this code? It duplicates the code in the configuration, and these models will follow OpenAI changes later.

```swift
enum OpenAIModels: String, CaseIterable {
    case gpt3_5_turbo_0125 = "gpt-3.5-turbo-0125"
    case gpt4_0125_preview = "gpt-4-0125-preview"
}
```

phlpsong: You're right, I have rewritten the `OpenAIService` to reuse this enum type.

tisfeng: ok

tisfeng (approved):
Looks good.

tisfeng:
Please review this PR.

Jerry23011 (commented):
This popup looks like one with a mouse right click, is it possible that we can have a fixed position for the popup?
截屏2024-03-21 14 00 23

phlpsong:
Could you share some examples here? I'm not quite sure what you meant by the fixed position.

Jerry23011:
Certainly, the popups in settings have a fixed position no matter where I click on them. Please see recording below

https://github.com/tisfeng/Easydict/assets/89069957/06b75f98-cf8f-4839-9e5b-321616abcd9c

tisfeng:
Yes, the NSMenu should be displayed in a fixed position, e.g. clicking on the bottom left corner of the view, which looks best.

image

tisfeng:
I didn't look at the detailed reason why the atLocation coordinates would be drifting, it seems a little strange.

Also, I noticed that NSMenu is used in other places, such as `EZSelectLanguageButton`,  we can write a Category for NSMenu, making it convenient to use to show NSMenu at the bottom left of a view.

phlpsong:
Thanks, I got you. 👍

But personally, the Picker-like component is not suitable in this case, popup the menu from the left corner looks better.

tisfeng:
ok, we don't need to change the system Picker, just the NSMenu.

tisfeng (approved):

Jerry23011 (approved):
LGTM

Issue #471, opened 2024-03-25
🐞 Bug Report: Main window can not again after press escape while input query word


phlpsong:
### Please confirm the following:

- [X] Be sure to check out [FAQ](https://github.com/tisfeng/Easydict/wiki/FAQ)
- [X] I have carefully read the [README](https://github.com/tisfeng/Easydict#readme)
- [X] I have searched through the [issues](https://github.com/tisfeng/Easydict/issues) page but couldn't find a solution.(Including issue that has been closed)
- [X] Easydict has been upgraded to the [latest version](https://github.com/tisfeng/Easydict/releases)
- [X] I understand and agree to the above, and understand that the project maintainer has limited energy, **issues that do not follow the rules may be ignored or closed directly**

### Problem description

Can not open Main query window again after press escape key while input query word

### Is the issue consistently reproducible?

Reproducible

### Steps to Reproduce

1. Open app with Main window shown
2. Focus on the input view on top of query window
3. Main window hide and cannot open again

### Screenshot

N/A

### Expected result

the Main window should be shown again 

### Device type and OS version

macOS 14.2.1

### Are you willing to submit a PR to fix this issue?

- [X] I'm willing to submit a PR

tisfeng:
[2.7.0](https://github.com/tisfeng/Easydict/releases/tag/2.7.0) 版本已修复。

PR #472, opened 2024-03-25
Main window cannot open after escape key pressed

phlpsong:
close https://github.com/tisfeng/Easydict/issues/471


tisfeng:
After this PR, we now no longer need to destroy the main window in most cases, such as by clicking the close button in the upper left corner, or the escape key.

To simplify the logic, I think we can remove the `closeFloatingWindowExceptMain` method and use the `closeFloatingWindow` method, just like fixedWindow, and change the main window to not release when closed `mainWindow. releasedWhenClosed = NO`.

I simply tested it and it works.

https://github.com/tisfeng/Easydict/blob/e5ce53bc20ad6b34d04dff62a05fa2d6a361818c/Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m#L196

```objc
- (EZFixedQueryWindow *)fixedWindow {
    if (!_fixedWindow) {
        _fixedWindow = [EZFixedQueryWindow shared];
        _fixedWindow.releasedWhenClosed = NO;
    }
    return _fixedWindow;
}
```

phlpsong:
You are right. It's a better approach.

tisfeng:
@phlpsong  Sorry, I didn't consider it thoroughly before. In some cases, it is not necessary to close the main window, such as when the main window loses focus or when opening App links.

I have improved the code by adding `closeFloatingWindowIfNotPinned` and `closeFloatingWindowIfNotPinnedOrMain` methods to handle different situations.

The specific usage here is a bit complex, please test the effect of the code carefully.

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
```diff
@@ -912,31 +911,42 @@ - (void)closeFloatingWindow {
...
+}
+
+- (void)closeFloatingWindowIfNotPinned {
+    [self closeFloatingWindowIfNotPinnedAndExceptWindowType:EZWindowTypeNone];
+}
+
+- (void)closeFloatingWindowIfNotPinnedAndExceptWindowType:(EZWindowType)windowType {
+    BOOL isPinned = self.floatingWindow.isPin;
+    BOOL isTypeMatch = self.floatingWindow.windowType == windowType;
+    if (!isPinned && !isTypeMatch) {
```
phlpsong, line 948: I wonder if we need this `windowType` param here since we generally exclude the main type only for now.

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
```diff
@@ -912,31 +911,42 @@ - (void)closeFloatingWindow {
...
     
     if (![self currentShowingSettingsWindow]) {
         // recover last app.
         [self activeLastFrontmostApplication];
     }
     
-    if ([EZMainQueryWindow isAlive]) {
-        [self.mainWindow orderBack:nil];
+    /// !!!: Maybe floating window has closed after calling [self.floatingWindow close], so we need to check it.
+    if (self.floatingWindow == closingFloatingWindow) {
```
phlpsong, line 928: In what cases they are different?

phlpsong:
Looks good.

I have left some comments and would appreciate it if you could clarify them.

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
tisfeng: Yes, but sometimes we don't need to exclude the main window, it's here to reduce duplicate code.

```objc
- (void)closeFloatingWindowIfNotPinnedOrMain {
    [self closeFloatingWindowIfNotPinnedAndExceptWindowType:EZWindowTypeMain];
}

- (void)closeFloatingWindowIfNotPinned {
    [self closeFloatingWindowIfNotPinnedAndExceptWindowType:EZWindowTypeNone];
}

- (void)closeFloatingWindowIfNotPinnedAndExceptWindowType:(EZWindowType)windowType {
    BOOL isPinned = self.floatingWindow.isPin;
    BOOL isTypeMatch = self.floatingWindow.windowType == windowType;
    if (!isPinned && !isTypeMatch) {
        [self closeFloatingWindow];
    }
}
```

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
tisfeng: When enable main window, the fixed window is displayed for the first time, and then clicking on the blank space triggers `setDismissFixedWindowBlock` to close the window, which triggers `closeFloatingWindow` twice, the first time the floating window is the fixed window, the second time is the main window.

The second time is an exception, because the `floatingWindow` is dynamically changing according to the `floatingWindowType`, and its value may change in the middle.

image

tisfeng: Since closing a window causes the floatingWindowType to change, I've added a windowType parameter to specify that a window should be closed, avoiding the problem mentioned above.

```objc
- (void)windowDidResignKey:(NSNotification *)notification {
//    NSLog(@"windowDidResignKey: %@", self);
    
    // Close floating window when losing focus if it's not pinned or main window.
    [EZWindowManager.shared closeFloatingWindowIfNotPinned:self.windowType exceptWindowType:EZWindowTypeMain];
}
```

Easydict/Feature/ViewController/Window/WindowManager/EZWindowManager.m
phlpsong: ok, I get it. 

Could we move `Not Pinned` check to `closeFloatingWindow`, it looks like we should never close a pinned window in any case.

tisfeng: No, in some cases, even if it is a pinned window, we still need to be able to close it. 

For example, manually using the escape key or using shortcut keys to show/close the query window. You can refer to the invocation of the `closeFloatingWindow` method for details.

These are existing interaction logics that indicate users indeed want to manually close this window. I think we should not disrupt it.

phlpsong: Update some code, let me know what you think.

phlpsong: ok

tisfeng (approved):

Jerry23011 (approved):

AkaShark (approved):

PR #483, opened 2024-03-31
feat: add action to replace newline with space

tisfeng:
Close https://github.com/tisfeng/Easydict/issues/86#issuecomment-2028535304

image


tisfeng:
No, this action is usually only needed in some special cases, so I don't think we need to add an extra option.

Jerry23011 (approved):

PR #493, opened 2024-04-06
perf: enable to customize OpenAI models

tisfeng:
close https://github.com/tisfeng/Easydict/issues/489

image


Easydict/Swift/Feature/Configuration/Configuration+Defaults.swift
```diff
@@ -201,7 +201,11 @@ extension Defaults.Keys {
         default: OpenAIUsageStats.default
     )
     static let openAIEndPoint = Key<String?>(EZOpenAIEndPointKey)
-    static let openAIModel = Key<OpenAIModel>(EZOpenAIModelKey, default: .gpt3_5_turbo_0125)
+    static let openAIModel = Key<String>(EZOpenAIModelKey, default: OpenAIModel.gpt3_5_turbo.rawValue)
+    static let openAIAvailableModels = Key<String?>(
+        "EZOpenAIAvailableModelsKey",
```
phlpsong, line 206: It would be better to define a constant string like `EZCustomOpenAIModelssAvailableKey `

phlpsong:
https://github.com/tisfeng/Easydict/assets/103433299/be415ef6-f23c-48b9-aaee-76e3ed32a314

I found a issue when testing this pr. When switching models, the menu will occasionally pop up at the wrong cell. When you click on the custom OpenAI service, the menu will pop up in the OpenAI service cell, and vice versa may also occur.

 It looks like the service is wrong when popup the menu. You can check the below screenshots which I'm clicking the custom OpenAI service but the service type is OpenAI.
image


This also occurs in the dev branch. We need to take a look.

phlpsong (changes requested):

Easydict/Swift/Feature/Configuration/Configuration+Defaults.swift
tisfeng: Why? This value doesn't seem to be needed elsewhere.

phlpsong: The main reason is to keep consistent with others definitions like `EZOpenAIModelKey ` and `EZOpenAIAPIKey`.  It will be easy to manage them in a single place like `EZConstKey`. 

And do we need to support this key in schema parse read & write functions?

tisfeng: You're right, the current configuration of the OpenAI service is required to support command write modifications, so the `EZOpenAIAvailableModelsKey` needs to be open. I forgot about that.

But for some keys  that don't need to be external, such as `EZOpenAIValidModelsKey`, I think it's possible to just use literals?

phlpsong: I believe using literals is acceptable, but I would still recommend defining a constant like others. As mentioned above, this approach would make it easier to use if it needs to be external or remove it if not used anymore in the future.

tisfeng: ok

phlpsong:
I just noticed that you created `customOpenAIVaildModels ` and `openAIVaildModels `,  I would prefer to store them in one place and extract when needed since they are generally the same with the `AvailableModels`, it maybe redundant to store them.

Please let me know what you think.

tisfeng:
> I just noticed that you created `customOpenAIVaildModels ` and `openAIVaildModels `, I would prefer to store them in one place and extract when needed since they are generally the same with the `AvailableModels`, it maybe redundant to store them.
> 
> Please let me know what you think.

I know, this code is ugly, but I just wanted to simply finish this PR.

The main problem at the moment is that `OpenAIService + ConfigurableService` and `CustomOpenAIService + ConfigurableService` have too much similar and duplicated code.

This part of the code needs to be refactored and optimized, but I'm not very familiar with SwiftUI, so if you'd like, we can open a separate issue later to refactor it, which will make it easier for subsequent feature, like https://github.com/tisfeng/Easydict/issues/479 and https://github.com/tisfeng/Easydict/issues/492 .

phlpsong (approved):
LGTM

tisfeng:
Please continue to review, code duplication will be resolved by this issue https://github.com/tisfeng/Easydict/issues/497 .

phlpsong (approved):

PR #500, opened 2024-04-09
Fix tool tips

AkaShark:
close https://github.com/tisfeng/Easydict/issues/498


Easydict/objc/ViewController/View/Titlebar/EZTitlebar.m
```diff
@@ -307,7 +324,14 @@ - (EZOpenLinkButton *)eudicButton {
...
         eudicButton.image = [[NSImage imageNamed:@"Eudic"] resizeToSize:self.imageSize];
-        eudicButton.toolTip = [NSString stringWithFormat:@"%@, %@", NSLocalizedString(@"open_in_eudic", nil), @"⌘+⇧+⏎"];
+        NSString *shortcutStr = Configuration.shared.eudicDictShortcutString;
+        NSString *toolTipStr = @"";
+        if (shortcutStr.length != 0) {
+            toolTipStr = [NSString stringWithFormat:@"%@, %@", NSLocalizedString(@"open_in_eudic", nil), shortcutStr];
+        } else {
+            toolTipStr = [NSString stringWithFormat:@"%@", NSLocalizedString(@"open_in_eudic", nil)];
+        }
+        eudicButton.toolTip = toolTipStr;
```
tisfeng, line 334: Too much duplicate code, please write a method to optimize this code.

Easydict/objc/ViewController/Window/WindowManager/EZWindowManager.m
```diff
@@ -455,6 +455,12 @@ - (void)updateFloatingWindowType:(EZWindowType)floatingWindowType isShowing:(BOO
 //    NSLog(@"after floatingWindowTypeArray: %@", self.floatingWindowTypeArray);
 }
 
+- (void)updateWindowsTitlebar {
+    [self.mainWindow.titleBar updateButtonsToolTip];
```
tisfeng, line 459:
```suggestion
    [_mainWindow.titleBar updateButtonsToolTip];
```

Normally, we should use `_mainWindow` because the main window is turn off by default, so there's no need to create it.

tisfeng (changes requested):

Easydict/Swift/Feature/Configuration/Configuration+Defaults.swift
```diff
@@ -189,6 +189,35 @@ class DefaultsWrapper<T: Defaults.Serializable> {
...
+        var keyEquivalent = ""
+        var modifier = ""
+        if let key = keyCombo, key.doubledModifiers {
+            modifier = keyCombo?.keyEquivalentModifierMaskString ?? ""
+            keyEquivalent = keyCombo?.keyEquivalentModifierMaskString ?? ""
+        } else {
+            modifier = keyCombo?.keyEquivalentModifierMaskString ?? ""
+            keyEquivalent = keyCombo?.keyEquivalent ?? ""
+        }
+        return modifier + keyEquivalent
```
phlpsong, line 217:
```suggestion
        if let keyCombo, keyCombo.doubledModifiers {
            return keyCombo.keyEquivalentModifierMaskString + keyCombo.keyEquivalentModifierMaskString
        }
        return (keyCombo?.keyEquivalentModifierMaskString ?? "") + (keyCombo?.keyEquivalent ?? "")
```

Easydict/Swift/Feature/Configuration/Configuration.swift
```diff
@@ -588,6 +624,26 @@ extension Configuration {
...
+
+    fileprivate func didSetGoogleShortcut() {
+        let windowManager = EZWindowManager.shared()
+        windowManager.updateWindowsTitlebar()
+    }
+
+    fileprivate func didSetEudicDictShortcut() {
+        let windowManager = EZWindowManager.shared()
+        windowManager.updateWindowsTitlebar()
+    }
```
phlpsong, line 646: It looks like they are using the same logic, it would be better to replace them with a general function.

phlpsong (changes requested):

Easydict/Swift/Feature/Configuration/Configuration.swift
phlpsong: Actually, I mean we could remove the `didSetAppleDictShortcut ` and other 3 functions and use  `updateWindowTitlebar ` when receive Defaults value changes. What do you think?

phlpsong (dismissed):
LGTM!

Easydict/Swift/Feature/Configuration/Configuration.swift
```diff
@@ -419,6 +424,37 @@ class Configuration: NSObject {
...
+                }
+        )
+
+        cancellables.append(
+            Defaults.publisher(.eudicShortcut)
+                .removeDuplicates()
+                .sink { [weak self] _ in
+                    self?.updateWindowTitlebar()
+                }
+        )
```
tisfeng, line 457:
```suggestion
         Publishers.Merge4(
            Defaults.publisher(.pinShortcut),
            Defaults.publisher(.appleDictionaryShortcut),
            Defaults.publisher(.googleShortcut),
            Defaults.publisher(.eudicShortcut)
        )
        .removeDuplicates()
        .sink { [weak self] _ in
            self?.updateWindowTitlebar()
        }
        .store(in: &cancellables)
```

Too much duplicate code, can we optimize it like this?

tisfeng: Also, I noticed that `cancellables` is an Array type instead of a Set, why?

```swift
var cancellables: [AnyCancellable] = []
```

tisfeng: Can we use Set here? `store(in:` looks more concise than the nested `append`.

tisfeng (changes requested):

Easydict/Swift/Feature/Configuration/Configuration.swift
tisfeng: Also, I noticed that `cancellables` is an Array type instead of a Set, why?

```swift
var cancellables: [AnyCancellable] = []
```

Can we use Set here? `store(in:` looks more concise than the nested `append`.

tisfeng:
I improved the code, please continue to review.

Easydict/Swift/Feature/Configuration/Configuration.swift
```diff
@@ -162,263 +167,246 @@ class Configuration: NSObject {
...
 
-                    self?.didSetAppearance(newValue)
+        let shortcutKeys: [Defaults.Key] = [
+            .pinShortcut,
+            .appleDictionaryShortcut,
+            .googleShortcut,
+            .eudicShortcut,
+        ]
+        for key in shortcutKeys {
+            Defaults.publisher(key)
```
phlpsong, line 403: ` Defaults.publisher(keys:)` can be used in this case.

tisfeng: I tried it, but it doesn't seem to work with `removeDuplicates()`.

If you can fix it, please provide code suggestions.

phlpsong: Yes, not having this interface https://github.com/sindresorhus/Defaults/issues/43


You can use 'options' parameter and pass empty array to it to avoid receiving initial events. But it still publish two events after value changes.

If no other solution, it's fine to approve it.

tisfeng: ok，let's keep it.

phlpsong (approved):

tisfeng (approved):

PR #506, opened 2024-04-13
fix: launch app called windowDidResize

AkaShark:
closed https://github.com/tisfeng/Easydict/issues/504


AkaShark:
wait this still has some problems in other `publisher`

AkaShark:
looks `closeFloatingWindowIfNotPinnedOrMain` also called `windowDidResize` when launch and mouse click

phlpsong (approved):

Easydict/objc/ViewController/Window/WindowManager/EZWindowManager.m
```diff
@@ -921,6 +921,9 @@ - (void)closeFloatingWindowIfNotPinnedOrMain:(EZWindowType)windowType {
 }
 
 - (void)closeFloatingWindowIfNotPinned:(EZWindowType)windowType exceptWindowType:(EZWindowType)exceptWindowType {
+    if (_fixedWindow == nil && _miniWindow == nil) {
```
tisfeng, line 924: This solution is not good. This condition is very strange to be placed here. It is just to handle a special situation. 

In addition, it will cause the main window to not be closed.

tisfeng: I fixed it in 481191cfafadef65656c4df3a81c6aa27c22e38f

Easydict/Swift/Feature/Configuration/Configuration.swift
```diff
@@ -402,8 +402,8 @@ class Configuration: NSObject {
         for key in shortcutKeys {
             Defaults.publisher(key)
                 .removeDuplicates()
-                .sink { [weak self] _ in
-                    self?.updateWindowTitlebar()
+                .sink { _ in
+                    EZWindowManager.shared().updateWindowsTitlebarButtonsToolTip()
```
phlpsong, line 406: we can replace this likes below

```swift
Defaults.publisher(keys: shortcutKeys, options: [])
            .throttle(for: 0.5, scheduler: DispatchQueue.main, latest: true)
            .sink { [weak self] _ in
                EZWindowManager.shared().updateWindowsTitlebarButtonsToolTip()
            }
            .store(in: &cancellables)
```

phlpsong: https://github.com/tisfeng/Easydict/pull/500#discussion_r1560374095 We have discussed this logic before, this should fix the duplicate event issue.

tisfeng: ok, this code looks a little better.

The last question, does this time threshold 0.5 have any special meaning, or changing it to 0.4 is also okay?

phlpsong: The time between repeated events is very short, and 0.5 is a relatively safe value to filter them. Based on my test 0.1 also works well.

phlpsong (approved):

Easydict/Swift/Feature/Configuration/Configuration.swift
tisfeng: ok

PR #514, opened 2024-04-17
fix: stop button was gone when querying stream service

### LINKED ISSUE #510: dev: OpenAI 翻译的停止按钮不见了
stream 响应时，这里应该有一个暂停按钮的，之前版本有，后面不知道被谁干掉了 😑

image

tisfeng:
Fixed #510 , and 
- enable to stop Gemini stream

https://github.com/tisfeng/Easydict/assets/25194972/5a18a321-3e11-4501-85a2-e259c2d9b75b

### DIFF
--- a/Easydict/objc/ViewController/Window/BaseQueryWindow/EZBaseQueryWindow.m
+++ b/Easydict/objc/ViewController/Window/BaseQueryWindow/EZBaseQueryWindow.m
@@ -128,7 +128,7 @@
 }
 
 - (void)windowDidResize:(NSNotification *)aNotification {
-    NSLog(@"windowDidResize: %@, windowType: %ld", @(self.frame), self.windowType);
+//    NSLog(@"windowDidResize: %@, windowType: %ld", @(self.frame), self.windowType);
 
     [[EZLayoutManager shared] updateWindowFrame:self];
     
--- a/Easydict/App/Localizable.xcstrings
+++ b/Easydict/App/Localizable.xcstrings
@@ -629,7 +629,7 @@
         },
         "zh-Hans" : {
           "stringUnit" : {
-            "state" : "needs_review",
+            "state" : "translated",
             "value" : "内置 AI 翻译"
           }
         }
@@ -4280,6 +4280,22 @@
           "stringUnit" : {
             "state" : "translated",
             "value" : "单词分词"
+          }
+        }
+      }
+    },
+    "stop" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Stop"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "停止"
           }
         }
       }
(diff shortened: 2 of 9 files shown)


Easydict/objc/ViewController/Window/BaseQueryWindow/EZBaseQueryWindow.m
```diff
@@ -128,7 +128,7 @@ - (void)windowDidResignKey:(NSNotification *)notification {
 }
 
 - (void)windowDidResize:(NSNotification *)aNotification {
-    NSLog(@"windowDidResize: %@, windowType: %ld", @(self.frame), self.windowType);
+//    NSLog(@"windowDidResize: %@, windowType: %ld", @(self.frame), self.windowType);
```
phlpsong, line 131: I saw some places where we are using `print` and `NSLog` to print some info,  maybe we could consider using a uniform log method later, we already import packages like `CocoaLumberjack` and `swift-log`.

phlpsong (approved):

Jerry23011 (approved):
LGTM

Jerry23011 (changes requested):
截屏2024-04-18 08 17 36
It seems that Stop is not localized under Simplified Chinese

Easydict/objc/ViewController/Window/BaseQueryWindow/EZBaseQueryWindow.m
tisfeng: Very good idea!

`swift-log` looks like a library that comes with `CocoaLumberjack`, but I'm not very familiar with `CocoaLumberjack`, and I'm confused about the relationship between them.

We actually have a legacy logging module in our project, `MMLog`, which provides useful crash logs when users experience crashes.

However, I think the `MMLog` code is not very good, it uses a lot of macros, looks complicated, and is not easy to use with external APIs, e.g., it doesn't distinguish well between normal logs, file logs, warning logs, etc. In addition, it uses some deprecated methods, resulting in having to use `#pragma clang diagnostic ignored` to ignore warnings.

I've thought about optimizing this part of the code before, but I'm not familiar with `CocoaLumberjack` and modifying it would be too difficult for me.

If you are familiar with log, can you help optimize this code?

tisfeng:
> 截屏2024-04-18 08 17 36 It seems that Stop is not localized under Simplified Chinese when hovering over the stop streaming button

ok, fixed.

But this `toolTip` is not very useful at the moment, because the way streaming requests are displayed now is so bad that it will cause the UI to keep refreshing and the `toolTip` may not be visible.

In addition, the frequent refresh of streaming request will also lead to UI display bugs and high CPU usage, which I didn't consider when designed the program at the beginning, but it's hard to modify it at the moment, so we can only wait until later to rewrite it with SwiftUI directly 😥

Jerry23011:
looking forward to seeing the rewrite then

Jerry23011 (approved):

AkaShark (approved):

AkaShark:
LGTM

Easydict/objc/ViewController/Window/BaseQueryWindow/EZBaseQueryWindow.m
phlpsong: Sure, I checked the documentation, and it seems pretty easy to use. Replacing the existing log should not take long.

PR #515, opened 2024-04-18
perf: improve validation error message

tisfeng:
Previously, the service validation page would simply display an error message without detailed error information when something went wrong, which could be confusing for users.

In addition, I've improved the code so that the error message for getting the validation service is consistent with the query action.

image

### DIFF
--- a/Easydict/App/Localizable.xcstrings
+++ b/Easydict/App/Localizable.xcstrings
@@ -3192,13 +3192,13 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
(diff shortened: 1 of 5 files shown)


phlpsong (approved):

Jerry23011 (commented):
截屏2024-04-18 08 23 09
DeepL 似乎没有返回值，不知道是不是正常现象

tisfeng:
虽然这个错误信息看起来很奇怪（API key 错误为什么是 403 forbidden 🤔），但错误日志就是这样，代码应该没问题。

image

PR #516, opened 2024-04-18
perf: do not hide model button when loading

tisfeng:
https://github.com/tisfeng/Easydict/assets/25194972/9d09e48d-c5da-4c73-babd-c7aed26eafa0

### DIFF
--- a/Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/ServiceConfigurationCells.swift
+++ b/Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/ServiceConfigurationCells.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2024 izual. All rights reserved.
 //
 
+import Combine
 import Defaults
 import SwiftUI
 
@@ -42,10 +43,16 @@ struct ServiceConfigurationSecureInputCell: View {
 struct ServiceConfigurationInputCell: View {
     // MARK: Lifecycle
 
-    init(textFieldTitleKey: LocalizedStringKey, key: Defaults.Key<String?>, placeholder: LocalizedStringKey) {
+    init(
+        textFieldTitleKey: LocalizedStringKey,
+        key: Defaults.Key<String?>,
+        placeholder: LocalizedStringKey,
+        limitLength: Int = Int.max
+    ) {
         self.textFieldTitleKey = textFieldTitleKey
         self.placeholder = placeholder
         _value = .init(key)
+        self.limitLength = limitLength
     }
 
     // MARK: Internal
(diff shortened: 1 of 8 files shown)


phlpsong:
Looks good, but I found some layout problems need to take a look:

1. Model text switched to some other model name when pointer hover on  it

https://github.com/tisfeng/Easydict/assets/103433299/b1bb18cc-4628-482a-8516-0a1982102ecb

2. The warning image  has an overlap issue with the model selection button
Screenshot 2024-04-18 at 23 01 46

3. The loading icon is out of the bounds of the area and overlaps with the arrow image, it could be more obvious if service name is longer

https://github.com/tisfeng/Easydict/assets/103433299/e0d24ff0-f8c8-40a0-be05-0b877f6f4967

tisfeng:
All the issues have been resolved. 

I increased the minimum width of the window a bit, limited the length of custom OpenAI names, and also limited the maximum length of the service name label and model button.

image

Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/ServiceConfigurationCells.swift
```diff
@@ -54,9 +61,22 @@ struct ServiceConfigurationInputCell: View {
     let textFieldTitleKey: LocalizedStringKey
     let placeholder: LocalizedStringKey
 
+    var limitLength: Int
+
     var body: some View {
         TextField(textFieldTitleKey, text: $value ?? "", prompt: Text(placeholder))
             .padding(10.0)
+            .onReceive(Just(value)) { _ in
+                limit(limitLength)
```
tisfeng, line 70: This implementation doesn't work very well, it can display over the length when typing and then truncate.

I've tried the answers here [how-to-set-textfield-max-length](https://stackoverflow.com/questions/56476007/how-to-set-textfield-max-length) , but none of them work well, can you help look at this?

tisfeng:
Show service name label and model button width based on window width ratio, if window has enough width, they can be shown fully.

image

Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/ServiceConfigurationCells.swift
phlpsong: I couldn’t come up with anything else, either. 

We have constraints on the display in settings and query window, maybe it's fine to not limit this place.

phlpsong:
https://github.com/tisfeng/Easydict/assets/103433299/c5d9b56d-5587-493e-85c8-f9124b1cc581

This issue still exists if you press the modal name instead of hovering over it.

Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/ServiceConfigurationCells.swift
tisfeng: Yes, there are limits elsewhere, so it doesn't make sense to have too long inputs here, and I want the user to know that there is a limit to the length of inputs here.

If there is no good solution, leave it like this.

tisfeng:
ok, fixed.

`EZButton` is a control I wrote casually at that time. This is my first time writing a macOS application, and many codes were not very good 😥

Although the code is bad, there is no need to optimize it. Let's rewrite it with SwiftUI later.

phlpsong (approved):

PR #518, opened 2024-04-19
refactor: replace NSLog and print with DDLog

phlpsong:
close https://github.com/tisfeng/Easydict/issues/517

### DIFF
--- a/Easydict/objc/MMKit/Log/MMFileLogFormatter.m
+++ b/Easydict/objc/MMKit/Log/MMFileLogFormatter.m
@@ -41,16 +41,16 @@
             logLevel = @"❌";
             break;
         case DDLogFlagWarning:
-            logLevel = @"W";
+            logLevel = @"⚠️";
             break;
         case DDLogFlagInfo:
-            logLevel = @"I";
+            logLevel = @"ℹ️";
             break;
         case DDLogFlagDebug:
-            logLevel = @"D";
+            logLevel = @"Debug";
             break;
         default:
-            logLevel = @"V";
+            logLevel = @"Verbose";
             break;
     }
     return [NSString stringWithFormat:@"[%@ ● %@ ● %zd ● %@] %@ ● %@", [self stringFromDate:logMessage.timestamp], logMessage.fileName, logMessage.line, logLevel, logMessage.function, logMessage->_message];
--- a/Easydict/objc/MMKit/Log/MMFileLogFormatter.m
+++ b/Easydict/objc/MMKit/Log/MMFileLogFormatter.m
@@ -23,6 +23,14 @@
 
 @implementation MMFileLogFormatter
 
+#pragma mark - DDLogFormatter protocol
+
+- (NSString *)formatLogMessage:(DDLogMessage *)logMessage {
+    return [NSString stringWithFormat:@"[%@ ● %@ ● %zd ● %@] %@ ● %@", [self stringFromDate:logMessage.timestamp], logMessage.fileName, logMessage.line, [self logMessageEmoji:logMessage], logMessage.function, logMessage->_message];
+}
+
+#pragma mark -
+
 - (NSString *)stringFromDate:(NSDate *)date {
     // Single-threaded mode.
 
@@ -32,28 +40,6 @@
     }
 
     return [threadUnsafeDateFormatter stringFromDate:date];
-}
-
-- (NSString *)formatLogMessage:(DDLogMessage *)logMessage {
-    NSString *logLevel;
-    switch (logMessage->_flag) {
-        case DDLogFlagError:
-            logLevel = @"❌";
-            break;
-        case DDLogFlagWarning:
-            logLevel = @"⚠️";
-            break;
-        case DDLogFlagInfo:
-            logLevel = @"ℹ️";
-            break;
-        case DDLogFlagDebug:
-            logLevel = @"Debug";
-            break;
-        default:
-            logLevel = @"Verbose";
-            break;
-    }
-    return [NSString stringWithFormat:@"[%@ ● %@ ● %zd ● %@] %@ ● %@", [self stringFromDate:logMessage.timestamp], logMessage.fileName, logMessage.line, logLevel, logMessage.function, logMessage->_message];
 }
 
 @end
--- a/Easydict/objc/MMKit/Log/MMLog.h
+++ b/Easydict/objc/MMKit/Log/MMLog.h
@@ -13,7 +13,7 @@
  
  默认设置：
  * DEBUG 模式：
- *      输出 MMLogVerbose & MMLogInfo
+ *      输出 MMLogInfo & MMLogInfo
  *      同步打印
  * 非 DEBUG 模式:
  *      仅输出 MMLogInfo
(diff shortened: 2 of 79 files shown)


CanglongCl (approved):

AkaShark (approved):

tisfeng:
This PR looks good, but I noticed that you replaced all the `NSLog` with `MMLogInfo`, which might not be good.

The project actually has a macro before, `#define NSLog(frmt, ...) MMLogVerbose(frmt, ##__VA_ARGS__)`, 
https://github.com/tisfeng/Easydict/blob/e5ce53bc20ad6b34d04dff62a05fa2d6a361818c/Easydict/App/PrefixHeader.pch#L9
which has a very similar function, but this causes two problems:
1. sometimes we only want to print the log to the console, just for our debugging use, we don't want to write to the user log file.
2. sometimes the debug log content is very large, if all write to the log file, may lead to a large disk occupation.

Considering the above reasons, so I commented out this macro.

I would like you to add a `MMLog` to replace the system `NSLog` for console log printing only, but with better performance and log hierarchy management.

phlpsong:
We have this definition in `MMLog` which we could use to constraint log output in debug/release.

```objective-c
#if DEBUG
DDLogLevel MMDefaultLogLevel = DDLogLevelAll;
#else
DDLogLevel MMDefaultLogLevel = DDLogLevelInfo;
#endif
```

It means we only output `Error, warning and info logs` to file or console in release mode and all logs in debug mode.

Additionally, I have changed some of the `info` logs that I think are not very necessary to `verbose` and the file size configration.


```
typedef NS_ENUM(NSUInteger, DDLogLevel){
    /**
     *  No logs
     */
    DDLogLevelOff       = 0,

    /**
     *  Error logs only
     */
    DDLogLevelError     = (DDLogFlagError),

    /**
     *  Error and warning logs
     */
    DDLogLevelWarning   = (DDLogLevelError   | DDLogFlagWarning),

    /**
     *  Error, warning and info logs
     */
    DDLogLevelInfo      = (DDLogLevelWarning | DDLogFlagInfo),

    /**
     *  Error, warning, info and debug logs
     */
    DDLogLevelDebug     = (DDLogLevelInfo    | DDLogFlagDebug),

    /**
     *  Error, warning, info, debug and verbose logs
     */
    DDLogLevelVerbose   = (DDLogLevelDebug   | DDLogFlagVerbose),

    /**
     *  All logs (1...11111)
     */
    DDLogLevelAll       = NSUIntegerMax
};
```


Please let me know what you think.

Easydict/objc/MMKit/Log/MMLog.m
```diff
@@ -29,39 +29,27 @@ + (void)load {
...
-    }
+    // terminal, use os_log
+    [ddlog addLogger:[DDOSLogger sharedInstance]];
 
     // 文件输出
     MMFileLogFormatter *fileFormatter = [MMFileLogFormatter new];
     DDLogFileManagerDefault *fileManager = [[DDLogFileManagerDefault alloc] initWithLogsDirectory:[self logDirectoryWithName:name]];
-    fileManager.maximumNumberOfLogFiles = 200;
+    fileManager.maximumNumberOfLogFiles = 10;
     fileManager.logFilesDiskQuota = 20 * 1024 * 1024;
```
tisfeng, line 47: Sorry, I didn't notice before that there is already a file size limit here, in that case we don't need to limit the number of files.

Easydict/objc/MMKit/Log/MMLog.m
```diff
@@ -29,39 +29,27 @@ + (void)load {
...
-#pragma clang diagnostic pop
-    }
+    // terminal, use os_log
+    [ddlog addLogger:[DDOSLogger sharedInstance]];
 
     // 文件输出
     MMFileLogFormatter *fileFormatter = [MMFileLogFormatter new];
     DDLogFileManagerDefault *fileManager = [[DDLogFileManagerDefault alloc] initWithLogsDirectory:[self logDirectoryWithName:name]];
-    fileManager.maximumNumberOfLogFiles = 200;
+    fileManager.maximumNumberOfLogFiles = 10;
```
tisfeng, line 46:
```suggestion
    fileManager.maximumNumberOfLogFiles = 100;
```

Easydict/objc/MMKit/Log/MMLog.m
phlpsong: ok, changed it to 100.

Easydict/objc/MMKit/Log/MMFileLogFormatter.m
```diff
@@ -41,16 +41,16 @@ - (NSString *)formatLogMessage:(DDLogMessage *)logMessage {
...
-            logLevel = @"I";
+            logLevel = @"ℹ️";
             break;
         case DDLogFlagDebug:
-            logLevel = @"D";
+            logLevel = @"Debug";
             break;
         default:
-            logLevel = @"V";
+            logLevel = @"Verbose";
```
tisfeng, line 53: We should keep the formats of `MMFileLogFormatter` and `MMConsoleLogFormatter` consistent. There is some code duplication here, so we can optimize it.

Easydict/objc/MMKit/Log/MMLog.h
```diff
@@ -50,12 +50,15 @@ extern DDLogLevel MMDefaultLogLevel;
 extern BOOL MMDefaultLogAsyncEnabled;
 
 
-#define MMLogInfo(frmt, ...) LOG_MAYBE_TO_DDLOG([MMManagerForLog sharedDDLog], MMDefaultLogAsyncEnabled, MMDefaultLogLevel, DDLogFlagInfo, 0, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
 #define MMLogVerbose(frmt, ...) LOG_MAYBE_TO_DDLOG([MMManagerForLog sharedDDLog], MMDefaultLogAsyncEnabled, MMDefaultLogLevel, DDLogFlagVerbose, 0, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
```
tisfeng, line 53: As we already have the limitation on log file size and disk count, now we can use logs liberally, so can we remove `MMLogVerbose`? I think the three log types are enough: `MMLogInfo`, `MMLogWarn` and `MMLogError`.

phlpsong: I prefer to keep `verbose` in case we want to only output some info to the terminal.

Easydict/objc/MMKit/Log/MMFileLogFormatter.m
phlpsong: I tried but not familiar with OC implementation, could you help on this optimization?

tisfeng:
I added `MMLog` for printing logs in debug mode without writing to the user log file, for logs that need to be written to the user file I can use `MMLogInfo`.

```objc
#define NSLog(frmt, ...) MMLog(frmt, ##__VA_ARGS__)

#define MMLog(frmt, ...) LOG_MAYBE_TO_DDLOG([MMManagerForLog sharedDDLog], MMDefaultLogAsyncEnabled, MMDefaultLogLevel, DDLogFlagVerbose, 0, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
```

Easydict/objc/MMKit/Log/MMLog.h
phlpsong: To change all of the NSLog to `info` may have a problem that some useless info logs are written into the file.  Keeping the `verbose` may be helpful if we only output some info to the terminal.

tisfeng: Ok, I have kept the `verbose`, just changed it to `MMLog`.

You are right, I just removed this macro.

```
#define NSLog(frmt, ...) MMLog(frmt, ##__VA_ARGS__)
```

Easydict/objc/MMKit/Log/MMFileLogFormatter.m
tisfeng: ok, I will do this later.

Easydict/objc/MMKit/Log/MMLog.h
phlpsong: Ok, 'NSLog' may not a good idea, then we have two kinds of log interface styles. 

'MMLog' has the same problem IMHO, compare with 'MMLog'/'MMLogInfo'/'MMLogError', I think they are naming in different styles, maybe confused when using them.

tisfeng: That is my intention, I want to differentiate `MMLog` from `MMLogInfo` and `MMLogError`, `MMLog` is only used in debug mode.

The `verbose` naming is not as clear as `info`, `error,`  it reminds people of the `--verbose` parameter in the command line. Anyway, I can't tell its meaning at a glance without checking the documentation.

tisfeng: One of the simplest ways to understand it is to use `MMLog` just like `NSLog`.

tisfeng:
I have a question as to why the Type identifier for `MMLogError` in the Xcode console is a yellow warning ⚠️ instead of red ❗.

Also, the Type identifiers of `MMLogInfo` and `MMLogWarn` are the same, both are blue, which is strange.

image

image

tisfeng:
ok, I see. `DDLogFlagWarning` and `DDLogFlagInfo` are actually the same type at the system level, they're just categorized differently by `CocoaLumberjack`.

```objc
- (void)logMessage:(DDLogMessage *)logMessage {
    // Skip captured log messages
    if ([logMessage->_fileName isEqualToString:@"DDASLLogCapture"]) {
        return;
    }

    if (@available(iOS 10.0, macOS 10.12, tvOS 10.0, watchOS 3.0, *)) {
        NSString * message = _logFormatter ? [_logFormatter formatLogMessage:logMessage] : logMessage->_message;
        if (message != nil) {
            const char *msg = [message UTF8String];
            __auto_type logger = [self logger];
            switch (logMessage->_flag) {
                case DDLogFlagError  :
                    os_log_error(logger, "%{public}s", msg);
                    break;
                case DDLogFlagWarning:
                case DDLogFlagInfo   :
                    os_log_info(logger, "%{public}s", msg);
                    break;
                case DDLogFlagDebug  :
                case DDLogFlagVerbose:
                default              :
                    os_log_debug(logger, "%{public}s", msg);
                    break;
            }
        }
    }
}
```

AkaShark (approved):

Easydict/objc/MMKit/Log/MMLog.h
phlpsong: OK

Issue #528, opened 2024-05-01
updating to 2.7 forces to change the app language to chinese and there's no way to change it back to english.


hyperphysin:
### Please confirm the following:

- [X] Be sure to check out [FAQ](https://github.com/tisfeng/Easydict/wiki/FAQ)
- [X] I have carefully read the [README](https://github.com/tisfeng/Easydict#readme)
- [X] I have searched through the [issues](https://github.com/tisfeng/Easydict/issues) page but couldn't find a solution.(Including issue that has been closed)
- [X] Easydict has been upgraded to the [latest version](https://github.com/tisfeng/Easydict/releases)
- [X] I understand and agree to the above, and understand that the project maintainer has limited energy, **issues that do not follow the rules may be ignored or closed directly**

### Problem description

updating to 2.7 forces to change the app language to chinese and there's no way to change it back to english.

### Is the issue consistently reproducible?

Reproducible

### Steps to Reproduce

just update it

### Screenshot

![Screenshot 2024-05-01 at 11 10 09 AM](https://github.com/tisfeng/Easydict/assets/43456637/1490a461-89e0-4bf5-b787-35640c537c17)

### Expected result

I want it to be in english!

### Device type and OS version

apple m2 air, sonoma

### Are you willing to submit a PR to fix this issue?

- [ ] I'm willing to submit a PR

tisfeng:
Please try to change app language manually.

<img width="820" alt="image" src="https://github.com/tisfeng/Easydict/assets/25194972/ffe86e2c-ec6b-4121-a94f-b82af7fb0a38">

tisfeng:
@choykarl  If you have time, please take a look at this issue and see if it modifies the user's App language?

thesla13:
Hey, @tisfeng! I have the same issue @choykarl does.

The thing is the settings page looks nothing like the one in your screenshot. It never did look that way for me in previous versions either. It always looked like this:

![snap 2024 05 01 07-06-25](https://github.com/tisfeng/Easydict/assets/55979516/302eb976-2506-44ba-942b-bf3521432913)

And there doesn't seem to be an option to change the interface language. In your screenshot it's directly above the appearance setting, while there is nothing above the "外观" setting in the layout I happen to have.

I am very much confused, honestly ((

tisfeng:
We rewrote the settings page using SwiftUI and added many new features, such as app language switching, but this part of SwiftUI functionality requires macOS 13.0+.

[...]

PR #530, opened 2024-05-02
fix: improve local language

choykarl:
#528

### DIFF
--- a/Easydict/Swift/Feature/Localization/I18nHelper.swift
+++ b/Easydict/Swift/Feature/Localization/I18nHelper.swift
@@ -25,12 +25,20 @@ class I18nHelper: NSObject {
     }
 
     var localizeCode: String {
-        UserDefaults.standard.string(forKey: languagePreferenceLocalKey) ?? LanguageState.LanguageType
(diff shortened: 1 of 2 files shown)


Easydict/Swift/Feature/Localization/I18nHelper.swift
```diff
@@ -25,12 +25,20 @@ class I18nHelper: NSObject {
     }
 
     var localizeCode: String {
-        UserDefaults.standard.string(forKey: languagePreferenceLocalKey) ?? LanguageState.LanguageType
-            .simplifiedChinese.rawValue
+        guard let code = UserDefaults.standard.string(forKey: languagePreferenceLocalKey) else {
```
phlpsong, line 28: Looks like 'if let' is more suit for this case here.

FYI： https://www.hackingwithswift.com/quick-start/understanding-swift/when-to-use-guard-let-rather-than-if-let

phlpsong (changes requested):

phlpsong (approved):
LGTM

Jerry23011 (approved):

PR #532, opened 2024-05-02
Rewrite About view

Jerry23011:
Closes #449 

## Preview
截屏2024-05-03 17 13 22
截屏2024-05-03 17 13 28

## Known Issues
- The copyright info is not localized since it's directly extracted from the Info.plist
- ~~There seem to be no resize animations while switching from other tabs~~
- ~~The about view of the main app is still yet to be implemented, I don't know how to write the objc code for AppDelegate~~

PR #535, opened 2024-05-04
perf(UI): add Confetti effets to about views

Jerry23011:
### Preview

https://github.com/tisfeng/Easydict/assets/89069957/bf9b0590-e6f6-41f4-817c-d88b527f06ae

https://github.com/tisfeng/Easydict/assets/89069957/1f718efa-1aad-43bf-87ed-3dee0047203f


Jerry23011:
Oops, did it fail because of the minimum targets change? I could build it successfully on my Xcode

tisfeng:
It's strange, it looks like you've changed the layout?

## Now
image

## Before  

https://github.com/tisfeng/Easydict/pull/532#issuecomment-2093237751

image

Jerry23011:
Yes, I did add a ZStack so that the effect could work. Let me check it

Jerry23011:
I adjusted the constraints and now it should look the same with the original one.
The upper bound of the title varies by one pixel and the rest are exactly the same
## Now
Updated

## Before
Original

tisfeng:
ok, looks good.

tisfeng:
I think the layout can be optimized.

image

Easydict.xcodeproj/project.pbxproj
```diff
@@ -3297,7 +3301,7 @@
 				GCC_GENERATE_DEBUGGING_SYMBOLS = YES;
 				GCC_PREFIX_HEADER = "$(SRCROOT)/Easydict/App/PrefixHeader.pch";
 				GENERATE_INFOPLIST_FILE = YES;
-				MACOSX_DEPLOYMENT_TARGET = 11.0;
+				MACOSX_DEPLOYMENT_TARGET = 13.0;
```
tisfeng, line 3304: 稍后会发布一个小的修复版本，暂时不改最低支持版本，这里修改到了 macOS 13，因此这个 PR 放着等一会再合并。

Easydict/Swift/View/SettingView/Tabs/TabView/AboutTab.swift
```diff
@@ -27,50 +28,81 @@ struct AboutTab: View {
...
+                            .shadow(color: .gray, radius: 1, x: 0, y: 0.8)
+                            .padding(.bottom, 33)
+                            .padding(.leading, 16)
+                            .padding(.trailing, 16)
+                            .onTapGesture {
+                                proxy.move(
+                                    to: CGPoint(
+                                        // Unify x-point for Settings and Menubar About view
+                                        x: (geometry.size.width / 2) - 159.5,
+                                        y: 70
```
tisfeng, line 61: Can this coordinate be dynamically calculated rather than hardcoded? 

If we need to modify the UI layout later, will this coordinate need to be adjusted accordingly?

tisfeng (changes requested):

Jerry23011:
> I think the layout can be optimized.
> 
> image

It's now equal, the 1 px difference at the bottom is caused by the shadow effect
image

Easydict/Swift/View/SettingView/Tabs/TabView/AboutTab.swift
Jerry23011: The sample code actually provides dynamic calculation
```swift
.onTapGesture { location in
    proxy.move(to: location)
    proxy.burst()
}
```
However, in our case, we're adding confetti to two views with difference window sizes, so I haven't found a solution yet.

https://github.com/tisfeng/Easydict/assets/89069957/ebc71427-2144-49cc-b8b2-1ac6a48ccc32

Jerry23011: One possible calculation is to find the relationship between the **width & heigh of the window** and **x & y**

We will be able to calculate the location of the logo and tap gesture location based on the window size

tisfeng:
I suddenly realized that our setting window can be resized, which can cause some strange coordinate issues.

The size of the setting window is carefully designed by us, and resizing it will affect the UI of the configuration. 

I think we should prohibit resizing the window, what do you think?


https://github.com/tisfeng/Easydict/assets/25194972/c25df66d-e8a0-4eaf-9e44-42f44fc9914d

Easydict/Swift/View/SettingView/Tabs/TabView/AboutTab.swift
tisfeng: ok

Jerry23011:
> I suddenly realized that our setting window can be resized, which can cause some strange coordinate issues.
> 
> The size of the setting window is carefully designed by us, and resizing it will affect the UI of the configuration.
> 
> I think we should prohibit resizing the window, what do you think?

I think that would be a great approach -- most apps from Apple (like what I'm using now, Safari) don't have resizable Settings window.

Jerry23011:
> Just out of curiosity, may I ask how the note the drawing is done. Like 117px is auto?

Yes it’s auto-calculated by the screenshot app I use: shottr.cc

AkaShark:
> > Just out of curiosity, may I ask how the note the drawing is done. Like 117px is auto?
> 
> Yes it’s auto-calculated by the screenshot app I use: [shottr.cc](shottr.cc)

Thank you, it is indeed a useful tool.

tisfeng (approved):

PR #537, opened 2024-05-05
Service enable not work well issue

phlpsong:
#536 

Not sure if this issue is caused by the translation/sentence/dictionary option being disabled, you could try this and give me feedback if this issue still exists.

[screen-capture.webm](https://github.com/tisfeng/Easydict/assets/103433299/55edc690-e830-4050-9646-b7db2ed02894)

### DIFF
--- a/Easydict/Swift/View/SettingView/Tabs/TabView/ServiceTab.swift
+++ b/Easydict/Swift/View/SettingView/Tabs/TabView/ServiceTab.swift
@@ -215,12 +215,6 @@ private struct ServiceItemView: View {
             }
             EZLocalStorage.shared().setService(service, windowType: viewModel.windowType)
             viewModel.postUpdateServiceNotification()
-            // check enable state in case of state not changed while saving
-            service.enabled = EZLocalStorage.shared().service(
-                service.serviceType(),
-                windowType: viewModel.windowType
-            ).enabled
-            serviceItemViewModel.isEnable = service.enabled
         }
         .toggleStyle(.switch)
         .controlSize(.small)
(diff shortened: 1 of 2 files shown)


Jerry23011:
I'm able to reproduce the issue on release v2.7.1, but the issue no longer appears on this PR.

I'm not sure if this PR caused a new issue of not being able to toogle on the service for several attempts. It seems to appear only on Built-in AI Translation service


https://github.com/tisfeng/Easydict/assets/89069957/d8dc1341-021c-4c19-960c-919897058a9a

phlpsong:
Appreciate for your feedback. 

Could u check if you disable all options in configuration? 

Enable state will not changed if user disable all the options like translation/sentence/dictionary. So I think it's an expected behavior no options enabled.

Jerry23011:
> Could u check if you disable all options in configuration?

Yes indeed, now it works, thanks

Jerry23011 (dismissed):

AkaShark (approved):

tisfeng:
I thought about it, and I feel that the enabled query types should not affect whether the service is enabled or not, as it may confuse users, and intuitively they should be independent.

So I changed the code 05ed2786 , please let me know what you think.

Jerry23011:
This is more convenient for new users, but would it be better if Translation query is toggled on the same time when the service is toggled?

Sample use case:
1. The service Built-in Translation is disabled, and all three query type are disabled
2. Toggle on Built-in Translation
3. Go into Built-in Translation and find that all three query types are disabled.

In this case, would it be nice if translation query is toggled on at the same time so that users know that it's the default query type?

tisfeng:
I think it would be better to keep the service enabled or not independent of the query type. Whether the service is enabled determines whether it is displayed in the query window, and the query type determines the query action.

Though your suggestion is good, it associates the enabling of the service with the query type, which could cause some cognitive complexity. 

I prefer to let users decide to turn on or off these query types, giving users freedom.

I'd like to try to keep it simple, keep the application simple to configure and use, and easy to understand.

phlpsong:
Looks good, it's better and more simple and will be less confusing for the user.

Jerry23011:
Yes that's true, LGTM

Jerry23011 (approved):

Issue #315, opened 2024-01-07
🐞 Bug Report: Empty Copy Beep not work


felixadlen:
### Please confirm the following:

- [X] I have carefully read the [README](https://github.com/tisfeng/Easydict#readme)
- [X] I have searched through the [issues](https://github.com/tisfeng/Easydict/issues) page but couldn't find a solution.(Including issue that has been closed)
- [X] Easydict has been upgraded to the [latest version](https://github.com/tisfeng/Easydict/releases)
- [X] I understand and agree to the above, and understand that the project maintainer has limited energy, **issues that do not follow the rules may be ignored or closed directly**

### Problem description

Triggering "Select Text Translate" command with a keyboard shortcut adds a Empty Copy Beep, if no text is selected. 
If text is selected, no beep.

Latest version (2.5.0) and I've tried toggle the "Disable Empty Copy Beep" but the beep is always there if no text is selected.

I

### Reproduction steps

Triggering "Select Text Translate" command with a keyboard shortcut adds a Empty Copy Beep, if no text is selected. 
If text is selected, no beep.

Latest version (2.5.0) and I've tried toggle the "Disable Empty Copy Beep" but the beep is always there if no text is selected.

### Screenshot

<img width="754" alt="image" src="https://github.com/tisfeng/Easydict/assets/56875180/73baf34d-357c-4fef-babe-a92253b1a515">

### Expected result

No audio beep when triggering the shortcut when no text is selected

### Device type and OS version

Apple M2, macOs 14.1.2

### Are you willing to submit a PR to fix this issue?

- [ ] I'm willing to submit a PR

tisfeng:
感谢你的反馈，最近版本我们并没有修改禁止提示音相关的代码，这个问题很可能一直存在。

最开始由于这个问题比较复杂 https://github.com/tisfeng/Easydict/issues/111 ，目前没有很好的解决办法，所以我们才提供一个 'Disable Empty Copy Beep' 选项，但这个选项可能会导致播放音乐卡顿  https://github.com/tisfeng/Easydict/issues/83 ，因此当前只有在没有播放声音的情况才会真正启用。

更糟糕的是，目前判断电脑是否有播放声音功能使用的是一个私有方法，这个方法不可靠，有时候失效，这种情况下  'Disable Empty Copy Beep' 会不工作。

tisfeng:
It's only sometimes that this problem occurs, at which point you can try playing-pausing the music to see if you can temporarily fix the problem.

[...]

PR #539, opened 2024-05-06
fix: turn off the alert volume instead of system volume

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/315 , #111, #83 , now Easydict won't interrupt music playback or make strange alert sounds.

Since we can modify the alert volume individually, so it doesn't affect the system volume, so I removed the `Disable Empty Copy Beep` option.


Jerry23011 (approved):
LGTM. This feature is working as expected by observing the alert volume in System Settings

AkaShark (approved):

PR #543, opened 2024-05-09
fix: remove fake window which the trick way to open setting

AkaShark:
closed https://github.com/tisfeng/Easydict/issues/528#issuecomment-2089805673

### DIFF
--- a/Easydict/Swift/View/MenuItemView.swift
+++ b/Easydict/Swift/View/MenuItemView.swift
@@ -67,18 +67,9 @@ struct MenuItemView: View {
                 latestVersion = version
             }
         }
-        .openSettingsAccess()
-        .onReceive(NotificationCenter.default.publisher(for: Notification.Name.openSettings, object: nil)) { _ in
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.05) {
-                // calling `openSettings` immediately doesn't work so wait a quick moment
-                try? openSettings()
-            }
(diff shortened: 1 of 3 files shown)


Easydict/Swift/View/MenuItemView.swift
```diff
@@ -67,10 +67,19 @@ struct MenuItemView: View {
                 latestVersion = version
             }
         }
+        .openSettingsAccess()
```
phlpsong, line 70: According to the doc of `SettingsAcess`, this should not used in menu content.

> If using a menu-based MenuBarExtra, do not apply openSettingsAccess() to the menu content. openSettings() cannot be used there due to limitations of SwiftUI.

https://github.com/orchetect/SettingsAccess?tab=readme-ov-file#2-use-in-a-menubarextra-menu

phlpsong:
Screenshot 2024-05-09 at 10 52 16

The menu UI looks wired after this change.

AkaShark:
> Screenshot 2024-05-09 at 10 52 16
> The menu UI looks wired after this change.

You are right, let me change the implementation

phlpsong:
This works well for me.👍

phlpsong (dismissed):

tisfeng (approved):
Looks good.

phlpsong (approved):

Jerry23011 (approved):

PR #548, opened 2024-05-09
perf: throttle the stream completion to update UI

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/541

When I designed this app, I didn't realize that there would be a stream service like OpenAI behind it, and its frequent refreshing of the TableView would lead to high CUP usage and even UI lagging, so I added a `Throttler` to limit the frequency of refreshing.

While this approach doesn't fundamentally solve the problem, it can greatly alleviate it.

Before we rewrote the query window using SwiftUI, this would be a temporary solution.


Jerry23011 (approved):
Works well on my machine

Jerry23011 (approved):

PR #561, opened 2024-05-22
refactor: add a base class LLMStreamService

tisfeng:
This PR has no functional changes. 

In order to better add and optimize other LLM services in the future, like #559, I've rewritten and optimized the LLM-related services. The main content includes:

- Added a LLM service base class `LLMStreamService`.
- Modified `BaseOpenAIService` to inherit from `LLMStreamService`.
- Refactored and modified `GeminiService` to inherit from `LLMStreamService`.
- Updated swiftlint.yml, and remove unnecessary `swiftlint:disable` code.


Easydict/Swift/Service/OpenAI/LLMStreamService.swift
```diff
@@ -0,0 +1,129 @@
...
+
+    // MARK: Internal
+
+    let throttler = Throttler()
+    var updateCompletion: ((EZQueryResult, Error?) -> ())?
+
+    var model = ""
+
+    var unsupportedLanguages: [Language] {
+        []
```
phlpsong, line 73: Could we change these default return values to `fatalError("Not implemented")`, that will remind us to override these in subclasses.

.swiftlint.yml
```diff
@@ -33,6 +34,7 @@ opt_in_rules:
 line_length:
   warning: 120
   ignores_comments: true
+  ignores_interpolated_strings: true
```
phlpsong, line 37: Nice to know this 👍

Easydict/Swift/Service/OpenAI/LLMStreamService.swift
tisfeng: I think it's fine to keep it that way, for most LLM services we don't need to specify unsupported languages for several reasons:

- LLM services usually don't explicitly tell us all the languages they support, unless we test them all, which is too much trouble.
- LLM services usually keep updating their versions, and the languages they support may change as well, making it difficult for us to keep up with it.
- Most importantly, LLM services are not very sensitive to language types, and sometimes it can get correct translation results even if it recognizes the wrong language type.

For example, here is a Latin sentence, which is incorrectly recognized as Dutch because we don't support that language in Easydict at the moment, but even so, most LLM services translate it correctly, which is great 😃

### [Mundus vult decipi, ergo decipiatur.](https://weibo.com/6827625527/Of2Rm0k0s)


image

### [Google Translate](https://translate.google.com/?sl=auto&tl=zh-CN&text=Mundus%20vult%20decipi%2C%20ergo%20decipiatur.&op=translate)
image

phlpsong: Ok, that sounds fine. What about the other functions like  `availableModels ` / `apiKey ` / 'endpoint', in case of we forget to implement them in subclass, maybe consider using `fatal()`.

tisfeng: Good idea, I have marked `model`,`availableModels`, `apiKey` and `endpoint` as must be overriden.

phlpsong (approved):

PR #562, opened 2024-05-23
Change minimum deployments to 13.0

AkaShark:
closed https://github.com/tisfeng/Easydict/issues/534


AkaShark:
This PR remove a lot of code. Please help me review it carefully.

tisfeng (changes requested):
Why so many `STALE`?

image

tisfeng:
I think this is probably caused by the removal of the old settings page. 

Therefore, we should also remove these no longer applicable localized strings.

AkaShark:
> I think this is probably caused by the removal of the old settings page.
> 
> Therefore, we should also remove these no longer applicable localized strings.

Yeah, I forgot it. I will remove the `STALE` localized strings soon.

Easydict/objc/Configuration/EZConfiguration.h
tisfeng, line 1: Can we remove `EZConfiguration` now , since we already have `Configuration.swift`?

Easydict/objc/StatusItem/EZMenuItemManager.h
```diff
@@ -9,7 +9,7 @@
 #import <Foundation/Foundation.h>
 
 NS_ASSUME_NONNULL_BEGIN
-
+// will deprecation in next release now this class only used to fetch Repo last version in SwiftUI version. In this case it should be removed and move the fetch func to a new class
 @interface EZMenuItemManager : NSObject
```
tisfeng, line 13: This class is currently only being used by `fetchRepoLatestVersion`. Can you remove the other unused code?

Or you can remove everything and simply rewrite the `fetchRepoLatestVersion` method in Swift.

Easydict/objc/ViewController/Window/WindowManager/EZWindowManager.m
```diff
@@ -897,7 +893,7 @@ - (void)closeWindowOrExitSreenshot {
         [Snip.shared stop];
     } else {
         [self closeFloatingWindow];
-        [EZPreferencesWindowController.shared close];
+//        [EZPreferencesWindowController.shared close];
```
tisfeng, line 896: Just remove it.
```suggestion
```

Easydict/objc/ViewController/Window/WindowManager/EZWindowManager.m
```diff
@@ -959,7 +955,7 @@ - (void)closeFloatingWindow:(EZWindowType)windowType {
     [floatingWindow close];
     
     if (![self currentShowingSettingsWindow]) {
-        // Recover last app.
+//        // Recover last app.
```
tisfeng, line 958: There's no need to modify this.
```suggestion
       // Recover last app.
```

Easydict/objc/StatusItem/EZMenuItemManager.h
AkaShark: I note this class,  but `EZConfiguration` has a lot of keys, so I think we should remove it in next PR, in next PR will remove `EZConfiguration` and so on. This will make PR is not more complicated.

Easydict/objc/Configuration/EZConfiguration.h
AkaShark: I note this class, but EZConfiguration has a lot of keys, so I think we should remove it in next PR, in next PR will remove EZConfiguration and so on. This will make PR is not more complicated.

Easydict/objc/StatusItem/EZMenuItemManager.h
AkaShark: The same, I don't want to make this PR too complicated, so I comment in this place would like to make changes in the next PR.

Easydict/objc/Configuration/EZConfiguration.h
tisfeng: ok, it's better to do it in another PR.

Easydict/objc/StatusItem/EZMenuItemManager.h
tisfeng: ok

AkaShark: @tisfeng How do you think of this idea, this PR I want to try not to be more complicated, only change the deployment version, and try to determine some to modify, in the next PR to improve them.

Easydict/objc/Configuration/EZConfiguration.m
```diff
@@ -362,19 +362,9 @@ - (void)setHideMenuBarIcon:(BOOL)hideMenuBarIcon {
     
     [NSUserDefaults mm_write:@(hideMenuBarIcon) forKey:kHideMenuBarIconKey];
     
-    if (!Configuration.shared.enableBetaNewApp) {
```
phlpsong, line 365: I wonder if we could remove `EZConfiguration` entirely, it looks we already use `Configration` instead now.

phlpsong (changes requested):

Easydict/objc/Configuration/EZConfiguration.m
AkaShark: Yeah, I will remove `EZConfiguration` in the next PR because it has a lot of keys. I want to remove it carefully.

phlpsong: OK

Easydict/App/Localizable.xcstrings
```diff
@@ -4231,7 +3729,7 @@
         "zh-Hans" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "静默截图 OCR"
+            "value" : "静默截图OCR"
```
tisfeng, line 3732: We already have this string in SwiftUI, so it's ok remove these old strings.

image

Easydict/App/Localizable.xcstrings
```diff
@@ -2660,6 +2187,7 @@
       }
     },
     "select_translate" : {
+      "extractionState" : "manual",
```
tisfeng, line 2190: I found some duplicated localized strings, in addition to the new and old menu bar titles, there are also global shortcuts in the settings page, we should remove the duplicated values.

It's also ok to remove them in another PR.

image

AkaShark: okay

tisfeng (approved):

phlpsong (approved):

PR #574, opened 2024-06-01
crash report

AkaShark:
closed https://github.com/tisfeng/Easydict/issues/553


AkaShark:
Move crash and Log initialize into @main func

tisfeng:
We can now remove the old app launch, such as `Main.storyboard` https://github.com/tisfeng/Easydict/issues/553#issuecomment-2110588570 , and there may be some other unused code, please also check later.

It's ok to do it in another PR.

tisfeng (approved):

PR #578, opened 2024-06-03
Localize copyright info in about view

Jerry23011:
Finally figured out how to localize this. Referred to Safari's locales for Chinese localization

The parameter `value` is the default value for the localized string key
截屏2024-06-04 01 00 06
截屏2024-06-04 01 04 20


Easydict/App/InfoPlist.xcstrings
```diff
@@ -0,0 +1,60 @@
+{
+  "sourceLanguage" : "en",
```
AkaShark, line 2: can we use `Localizable.xcstrings`, why add a new localize strings file

Jerry23011: We cando that for the About view since we're just extracting string from the catalog, but the app info in /Applications won't be able to show the localized result based on the system language (like what we have now)
截屏2024-06-04 22 45 12

In my understanding, Since apps still uses `Localizable.strings` under the hood when is compiled, it can only extract strings for Info.plist from `InfoPlist.strings` files. Here's the documentation on how to localize `plist` files

https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/AboutInformationPropertyListFiles.html

AkaShark (approved):

Easydict/App/InfoPlist.xcstrings
```diff
@@ -0,0 +1,60 @@
...
+        }
+      }
+    },
+    "NSHumanReadableCopyright" : {
+      "comment" : "Copyright (human-readable)",
+      "extractionState" : "extracted_with_value",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "new",
```
phlpsong, line 46: Shall we change it to `translated`?

tisfeng: I'm not sure why it's `new` instead of `translated` here, but Xcode doesn't warn about it, so it doesn't seem to be a problem.

`InfoPlist.xcstrings` looks a bit different from `Localizable.xcstrings`, I notice that all en states are `new`, zh-Hans state is `translated`.

image

tisfeng:
Here is a useful blog about Apple app localization https://fatbobman.com/zh/posts/localizationapp-1/

image

tisfeng:
Apple Docs https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LocalizingYourApp/LocalizingYourApp.html#//apple_ref/doc/uid/10000171i-CH5-SW21

image

Easydict/App/InfoPlist.xcstrings
tisfeng: I recreated the `InfoPlis`t and tested this situation, and found that the extractionState of `InfoPlist.xcstrings` is `extracted_with_value` by default, while the state is `new`.

 If the value of extractionState is `manual`, the state is `translated`, which is the case for `Localizable.xcstrings`.

image

Jerry23011:
> Here is a useful blog about Apple app localization https://fatbobman.com/zh/posts/localizationapp-1/

If we're ever adding system service one day, we'll need a third catalog 🥹
截屏2024-06-06 23 41 02

Easydict/App/InfoPlist.xcstrings
AkaShark: Do we need to change `extractionState` to `manual` or use the default value

AkaShark: It sounds like `.xcstrings` may have the same behavior, but do we need to change `extractionState` to `manual`, Xcode isn't warning about it.

tisfeng: It doesn't seem to have any impact, this may be a minor issue with Xcode.

If we want to change `extractionState` to `manual`, we need to open the file in source code mode, which is a bit tedious. I think we can leave it as it is.

tisfeng (approved):
Looks good.

Easydict/App/InfoPlist.xcstrings
AkaShark: Yeah, I think changing it may be a bit complicated

AkaShark:
LGTM

phlpsong (approved):

PR #583, opened 2024-06-07
refactor: improve chat prompt messages

tisfeng:
This is a refactoring PR. It does not contain content changes. The main changes are:

- Use new methods to generate prompt conversations: `chatMessage` and `chatMessagePair`.
- Add a public method `serviceChatMessageModels` to `LLMStreamService` to convert dict prompts to model prompts.


Jerry23011 (approved):
LGTM

PR #592, opened 2024-06-16
perf: try to use ScreenCaptureKit to request screen access

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/587

Refer to the code of [QuickRecorder](https://github.com/lihaoyun6/QuickRecorder/blob/a86e795f092a3f3109eac09ca6bbbfa564652b0e/QuickRecorder/SCContext.swift#L53) .


tisfeng:
Apple says this is a feature, not a bug, so we don't need to fix it.

PR #603, opened 2024-07-10
Feat/ocr error tips

AkaShark:
closed https://github.com/tisfeng/Easydict/issues/478


Easydict/Swift/Feature/Configuration/Configuration+Defaults.swift
```diff
@@ -96,6 +96,7 @@ extension Defaults.Keys {
     )
 
     static var disableTipsView = Key<Bool>("disableTipsViewKey", default: false)
+    static var enableYoudaoOCR = Key<Bool>("enableYoudaoOCR", default: true)
```
tisfeng, line 99:
```suggestion
    static var enableYoudaoOCR = Key<Bool>("enableYoudaoOCR", default: false)
```

tisfeng: For advanced options, it's usually better to turn them off by default.

tisfeng (approved):
Looks good.

PR #604, opened 2024-07-10
fix: miss subscribers if llm service is turned off

tisfeng:
Fix https://github.com/tisfeng/Easydict/pull/599#issuecomment-2212466914


Easydict/Swift/Utility/GlobalContext.swift
```diff
@@ -21,6 +21,12 @@ class GlobalContext: NSObject {
             updaterDelegate: updaterHelper,
             userDriverDelegate: userDriverHelper
         )
+
+        for service in services {
```
phlpsong, line 25: Could this logic able to move inside `StreamConfigurationView` instead?

tisfeng: No, we need to ensure that the LLM service is only subscribed once and remains unreleased, and it is appropriate to put it in the initialization method of this singleton.

phlpsong: OK, this sounds reasonable.

I just concern the 'GlobalContext' sounds not the right place to do this. Maybe we could improve this in the future.

phlpsong (approved):
Issue fixed, good work👍

Easydict/Swift/Utility/GlobalContext.swift
tisfeng: Yes, this code should be improved later.

PR #610, opened 2024-07-12
feat: change fixed window to be both default window

tisfeng:
To make Easydict easier to use and reduce confusion for new users, such as https://github.com/tisfeng/Easydict/issues/565, I have decided to set the `fixed` window as the default window, and both mouse selection and shortcut keys will use the fixed window by default.

This PR will not affect existing users, and they can still choose to use either or both types of windows according to preference in the settings.

### DIFF
--- a/Easydict/App/Localizable.xcstrings
+++ b/Easydict/App/Localizable.xcstrings
@@ -3208,6 +3208,22 @@
         }
       }
     },
+    "setting.general.pin_window_when_showing" : {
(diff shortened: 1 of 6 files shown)


phlpsong (approved):

AkaShark (approved):

tisfeng:
I just pushed a wrong commit,  already reverted it 🥲

AkaShark (approved):

PR #626, opened 2024-08-01
fix: improve update cell height animation

tisfeng:
Left is new animation, right is old.

https://github.com/user-attachments/assets/83e48d14-8227-401f-9444-d76846d4a91c


Jerry23011:
The window doesn't seem to have that animation. Is it possible to add the animation to the window frame as well for consistency?

https://github.com/user-attachments/assets/36478008-8c2f-481e-96b2-1f613c4ff33b

tisfeng:
I know there is this issue, but the UI animations here are very complex. I have tried, and modifying the window animations causes the tableView to shift, resulting in strange effects.

I am not familiar with animations on macOS, so if you have any knowledge about it, please help fix it.

If it cannot be fixed for now, we can merge it first.

https://github.com/user-attachments/assets/1c5bd80b-11ab-4ea6-9b56-1586da19b5ec

Jerry23011:
I honestly have no idea with objc/AppKit APIs, since the feature's working, LGTM.

tisfeng:
Ok, let's merge it first.

 If someone can fix this animation issue, please feel free to submit a PR 🥺

PR #632, opened 2024-08-07
fix: reduce copying action to avoid clipboard issues caused by Cmd + C

tisfeng:
Close https://github.com/tisfeng/Easydict/issues/608


tisfeng:
Please review 🥺

Easydict/App/Localizable.xcstrings
```diff
@@ -1140,13 +1140,13 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Force get selected text when auto get selected text failed (Experimental)"
+            "value" : "Allow to use Cmd + C to force to get selected text when regular method fails (experimental feature)"
```
phlpsong, line 1143: Could you make this sentence shorter? And I think current sentence is not particularly clear to users what issue it will fix.

phlpsong:
Can we make this change as a default behavior but not an option to the user?

Easydict/App/Localizable.xcstrings
tisfeng: ok, changed to `Allow to use force to get selected text (experimental feature)`.

Easydict/App/Localizable.xcstrings
```diff
@@ -1135,18 +1135,18 @@
...
       }
     },
-    "force_auto_get_selected_text" : {
+    "force_get_selected_text" : {
       "localizations" : {
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Force get selected text when auto get selected text failed (Experimental)"
+            "value" : "Allow to use force to get selected text (experimental feature)"
```
phlpsong: What about `Enable/Allow forced text selection (experimental feature)`

tisfeng:
> Can we make this change as a default behavior but not an option to the user? This copy issue is annoying and some of the users may not hard to find the root cause and enable this option.

No, I would like to keep this option disabled by default, as it may cause clipboard copying issues https://github.com/tisfeng/Easydict/issues/608 . 

I have modified the text description, and now it will clearly inform the user of this.

If the user manually enables this option, he should be responsible for it.

Easydict/App/Localizable.xcstrings
tisfeng: good, it's `Allow forced text selection (note that it may affect clipboard copying)` now.

phlpsong (approved):
LGTM

PR #658, opened 2024-08-25
fix: clear query not work when select translate text is empty

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/657


Jerry23011 (approved):
Works on my end. LGTM

AkaShark (approved):

phlpsong (approved):

PR #675, opened 2024-09-15
refactor: rewrite AppleScript with Swift

tisfeng:
I rewrite the `AppleScriptTask` class which runs AppleScript in Swift, and added two extensions `AppleScriptTask+Browser.swift` and `AppleScriptTask+System.swift`, used for running browser AppleScript and system AppleScript.

Currently, running AppleScript defaults to using `NSAppleScript` as it's slightly faster than `NSTask`.


Jerry23011 (approved):
LGTM

Easydict/Swift/Utility/Extensions/Data/Data+Extension.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  Data+Extension.swift
+//  iPoem
```
phlpsong, line 3: Maybe change to `Easydict`?

phlpsong (approved):

PR #679, opened 2024-09-17
feat: support custom prompt for custom openai service

tisfeng:
Close https://github.com/tisfeng/Easydict/issues/434

image


tisfeng:
Please review 🥺

tisfeng:
🫠

Jerry23011:
I'll review this today

Jerry23011:
Seems that I'm still getting the default system prompts after setting my own in settings?
```json
{
  "stream": true,
  "model": "gpt-3.5-turbo-0125",
  "temperature": 0,
  "messages": [
    {
      "role": "system",
      "content": "You are a translation expert proficient in various languages, focusing solely on translating text without interpretation. You accurately understand the meanings of proper nouns, idioms, metaphors, allusions, and other obscure words in sentences, translating them appropriately based on the context and language environment. The translation should be natural and fluent. Only return the translated text, without including redundant quotes or additional notes."
    },
    {
      "role": "user",
      "content": "Translate the following English text into Simplified-Chinese text: \"\"\"The stock market has now reached a plateau.\"\"\""
    },
 ```

截屏2024-09-22 15 33 29

**Comment by tisfeng** [comment]
It seems that you have not enabled "Enable Custom Prompt", and you are using the default translation prompt.

**Comment by Jerry23011** [comment]
OK it works now. Is it possible to expand the System prompt and User prompt sections only when users enabled it?

**Comment by tisfeng** [comment]
Good idea, I have changed the code, now it shows custom prompt TextEditor only enabled.

File: Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/TextEditorCell.swift [hunk 1]
```diff
@@ -17,72 +17,148 @@ struct TextEditorCell: View {
...
+
+    @ViewBuilder
+    private var footnoteView: some View {
+        if let footnote = footnote {
+            Text(footnote)
+                .font(.footnote)
+                .foregroundStyle(.gray)
+                .fixedSize(horizontal: false, vertical: true)
+                .lineSpacing(3)
+                .textSelection(.enabled)
```
phlpsong, line 84: This text view will stretch after selecting text

Screenshot 2024-09-24 at 21 30 44

Easydict/App/Localizable.xcstrings
```diff
@@ -6002,6 +6070,74 @@
...
+        "en-CA" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "System Prompt"
+          }
+        },
+        "sk" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "System Prompt"
```
phlpsong, line 6124: `System Prompt` needs translation.

Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/TextEditorCell.swift
phlpsong: It looks like this is caused by `.lineSpacing(3)`.

tisfeng: OK, it's been fixed.

But why does lineSpacing cause this issue? If I want to add lineSpacing, how can I avoid it?

Easydict/App/Localizable.xcstrings
tisfeng: I know, but since specialized words like `Prompt` are not easy to translate, I chose to keep both `System Prompt` and `User Prompt`.

Or do you think `系统 Prompt` and `用户 Prompt` are better?

tisfeng:
If we look at the TextEditor component in isolation, the VStack layout is good, but here, to maintain consistency with the overall style of the Form, HStack might be better.

I'm not sure, everyone what do you think?

### HStack


### VStack

phlpsong:
The main reason is that TextEditor and title Text cannot well aligned. Use vertical style will look more clear.

Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/TextEditorCell.swift
phlpsong: I have no idea of the root cause, maybe it's a SwiftUI bug.
Same issue here: https://stackoverflow.com/questions/76113476/swiftui-bug-text-view-is-changing-line-spacing-when-selected

Jerry23011:
> The main reason is that TextEditor and title Text cannot well aligned. Use vertical style will look more clear.

Vertical +1

Jerry23011:
> I tried to add a transition animation, but it looks a bit strange.
> 
> I'm not very familiar with SwiftUI animations, so if anyone is good at this, please help improve it.
> 
>  iShot_2024-09-24_20.30.37.mp4

I would say the latest change is pretty neat without the transition.

I thought of [DisclosureGroup](https://developer.apple.com/documentation/swiftui/disclosuregroup)'s animation but don't know how it's implemented.

Easydict/App/Localizable.xcstrings
phlpsong: `系统 Prompt` and `用户 Prompt` are better.

tisfeng: ok, translated.

tisfeng:
I've changed the title and TextEditor to a vertical layout.

I tried [DisclosureGroup](https://developer.apple.com/documentation/swiftui/disclosuregroup) but it doesn't feel right here, just keep it that way for now.

phlpsong (approved):
LGTM

tisfeng:
这个功能已经实现了，周末看情况发个新版本。

Issue #677, opened 2024-09-16
🐞 Bug Report: Apple Dict doesn't work in some languages


windrunner414:
### Please confirm the following:

- [X] Be sure to check out [FAQ](https://github.com/tisfeng/Easydict/wiki/FAQ)
- [X] I have carefully read the [README](https://github.com/tisfeng/Easydict#readme)
- [X] I have searched through the [issues](https://github.com/tisfeng/Easydict/issues) page but couldn't find a solution.(Including issue that has been closed)
- [X] Easydict has been upgraded to the [latest version](https://github.com/tisfeng/Easydict/releases)
- [X] I understand and agree to the above, and understand that the project maintainer has limited energy, **issues that do not follow the rules may be ignored or closed directly**

### Problem description

Apple Dict in Easydict returns no result when the query text contains hiragana/katakana (Japanese) or some Swedish letters (å, ä, ö). I only tested Japanese and Swedish, it's possible that all characters other than Chinese and Latin don't work.
For example, はなし:
超級クラウン中日辞典 / クラウン日中辞典 (the built-in one in apple dictionary):
<img width="578" alt="Screenshot 2024-09-16 at 3 45 49 AM" src="https://github.com/user-attachments/assets/b638e07b-0fd4-4518-a489-7c3b43a7fe1d">
Easydict:
<img width="472" alt="Screenshot 2024-09-16 at 3 46 21 AM" src="https://github.com/user-attachments/assets/6cbb33e2-89df-4b9b-9b59-e94be145d11d">

However, 話 will return the correct results
<img width="469" alt="Screenshot 2024-09-16 at 3 48 15 AM" src="https://github.com/user-attachments/assets/023b41f7-1495-48f0-b782-641c91c30af7">

### Is the issue consistently reproducible?

Reproducible

### Steps to Reproduce

1. Enabling 超級クラウン中日辞典 / クラウン日中辞典 in Apple Dictionary
2. Quering any word contains hiragana, e.g. はなし. Apple Dictionary will show the correct result but Easydict will say no result

### Screenshot

<img width="472" alt="Screenshot 2024-09-16 at 3 46 21 AM" src="https://github.com/user-attachments/assets/6cbb33e2-89df-4b9b-9b59-e94be145d11d">

### Runtime log(zip file)

Sorry there are too many menu icons on my laptop and the easydict one is not showing up, I'll try to get the log later

### Expected result

Easydict shows the correct result as Apple Dictionary does

### Device type and OS version

[...]

PR #681, opened 2024-09-18
fix: do not need to check text query type when querying apple dictionary

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/677

image


phlpsong (approved):
LGTM

PR #703, opened 2024-10-26
feat: add window configuration to allow hiding the text input and language selection bar

tisfeng:
Close https://github.com/tisfeng/Easydict/issues/700

image


Jerry23011:
Love the feature.

I'm wondering why the two toggles are placed in the Service settings.

After selecting a service, the toggles are "hidden" until clicking on an option in the top bar. This may be a bit inconvenient.

tisfeng:
I know, but I can't find a suitable place to put it.

Window configuration is only related to window types and involves multiple window types. I don’t want to add a separate tab, and at this moment, I noticed that it can reuse the UI of the Service configuration.

Perhaps the tab title can be changed to "Window Service" to show its relevance to both window and service configuration.

If you have a better UI design, please let me know.

Jerry23011:
> Window configuration is only related to the window type and involves multiple window types

I see, didn't realize the configs are separate for different window types.

Then is it possible to add a "go back" or "go home" button somewhere to go back to the service settings. Or is there a way to deselect a service? For now I can only do that by switching a new window type, which is not very intuitive.

tisfeng:
Sounds good, I added a back button that returns from service configuration to window configuration when clicked.

image

Jerry23011 (approved):

PR #711, opened 2024-10-31
feat: support zh-hant for DeepL

tisfeng:
Close https://github.com/tisfeng/Easydict/issues/709

Because DeepL's Source language and Target language are different, Source languages do not support language variants, such as zh-hans, so we need to remove language variants when using Source languages.

[DeepL Docs](https://developers.deepl.com/docs/zh/resources/supported-languages#source-languages)

### DIFF
--- a/Easydict/objc/Service/DeepL/EZDeepLTranslate.m
+++ b/Easydict/objc/Service/DeepL/EZDeepLTranslate.m
@@ -27,7 +27,6 @@ static NSString *kDeepLTranslateURL = @"https://www.deepl.com/translator";
 
 - (instancetype)init {
     if (self = [super init]) {
-
     }
     return self;
 }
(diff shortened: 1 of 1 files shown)


Jerry23011 (approved):
LGTM

Easydict/objc/Service/DeepL/EZDeepLTranslate.m
```diff
@@ -154,59 +136,29 @@ - (BOOL)autoConvertTraditionalChinese {
...
     NSString *targetLangCode = [regionalVariant componentsSeparatedByString:@"-"].firstObject; // pt-PT, pt-BR
 
     NSString *url = @"https://"
-                    @"www2."
-                    @"deepl.com"
-                    @"/jsonrpc";
-    
+    @"www2."
+    @"deepl.com"
+    @"/jsonrpc";
```
phlpsong, line 152: I'm curious here why not make them a string like `@"https://www2.deepl.com/jsonrpc"` here, and I'm very surprised it's a valid declaration in objc.

phlpsong: Screenshot 2024-11-01 at 14 54 26

This is called "string literal concatenation" in C/Objective-C. Nice to know this 👨🏽‍💻

phlpsong (approved):

Easydict/objc/Service/DeepL/EZDeepLTranslate.m
tisfeng: Because we are using the free web API from DeepL here, I am worried they might cause me trouble, so I am using this concatenation method to avoid being searched.

This is a trick for splitting strings  [how-to-split-a-string-literal-across-multiple-lines-in-c-objective-c](https://stackoverflow.com/questions/797318/how-to-split-a-string-literal-across-multiple-lines-in-c-objective-c/797351#797351)

Issue #705, opened 2024-10-27
Missing strings for translation


Svec-Tomas:
### Please confirm the following:

- [X] Be sure to check out [FAQ](https://github.com/tisfeng/Easydict/wiki/FAQ)
- [X] I have carefully read the [README](https://github.com/tisfeng/Easydict#readme)
- [X] I have searched through the [issues](https://github.com/tisfeng/Easydict/issues) page but couldn't find a solution.(Including issue that has been closed)
- [X] Easydict has been upgraded to the [latest version](https://github.com/tisfeng/Easydict/releases)
- [X] I understand and agree to the above, and understand that the project maintainer has limited energy, **issues that do not follow the rules may be ignored or closed directly**

### Problem description

1. I want to change (update) the translation for this string, but can not locate it.
2. This string is in English, also I am not able to locate it in the file for translation.

### Is the issue consistently reproducible?

Reproducible

### Steps to Reproduce

Check for updates

### Screenshot

![CleanShot 2024-10-27 at 14 01 13@2x](https://github.com/user-attachments/assets/77edb1f3-884b-41a2-bcfc-1aad80b952e7)

### Runtime log(zip file)

[Easydict log 2024-10-27 14-05-08-281.zip](https://github.com/user-attachments/files/17534042/Easydict.log.2024-10-27.14-05-08-281.zip)

### Expected result

Strings being available for translation.

### Proposed solution (optional)

Adding the strings for translation.

### Device type and OS version

Apple MacBook 14 M1 Pro, MacOS 15.0.1

### Are you willing to submit a PR to fix this issue?

- [ ] I'm willing to submit a PR to fix this issue

tisfeng:
Thank you for your feedback.

We use [Sparkle](https://github.com/sparkle-project/Sparkle) for application updates, so this update prompt window belongs to Sparkle, not Easydict.

The string "Version History" has not been localized into SK, so the default English is used.

If you would like to translate it, perhaps you can submit an issue to their project.

<img width="1670" alt="image" src="https://github.com/user-attachments/assets/8d353d32-e410-4f45-8ad1-a4b011997771">

Svec-Tomas:
Ah, I didn't know that. I hadn't found that information about using Sparkle anywhere before. Okay, I'll translate Sparkle as well, so then we'll have completely translated Easydict into Slovak!

tisfeng:
Here is a similar issue https://github.com/sparkle-project/Sparkle/pull/1261

[...]

PR #716, opened 2024-11-05
feat: add acknowledgements for SPM and pod

tisfeng:
https://github.com/tisfeng/Easydict/issues/705#issuecomment-2452815710

https://github.com/user-attachments/assets/6761115c-9353-41d4-a783-ce422e07db5d


AkaShark (approved):
LGTM

Jerry23011 (commented):
Is it possible to unify the window title bar style for navigation and detail view?
I think it would look better if they all have the style of the detail window
截屏2024-11-05 11 14 17
截屏2024-11-05 11 14 20

phlpsong (approved):

tisfeng:
Yes, I understand your thoughts.

I tried it, but it's not easy to do that.

The outer window is created by us, but the detail window is generated by the system's [NavigationStack](https://developer.apple.com/documentation/swiftui/navigationstack). I haven't found a good way to modify the height or style of the window title bar, so it's hard to make them uniform.

If you have a solution, please let me know.

Jerry23011:
> Yes, I understand your thoughts.
> 
> I tried it, but it's not easy to do that.
> 
> The outer window is created by us, but the detail window is generated by the system's [NavigationStack](https://developer.apple.com/documentation/swiftui/navigationstack). I haven't found a good way to modify the height or style of the window title bar, so it's hard to make them uniform.
> 
> If you have a solution, please let me know.

I see I see.

I know you can add `.task` to the AcknowListView for some AppKit customisations on the window style, but in my experience it always shatters, so I feel it's better to keep it this way.

Kyle-Ye:
There is some shiny new Windows APIs available on macOS 15+ which should help solve the issue.

eg.
- https://github.com/Kyle-Ye/MenuHelper/commit/d15174606d6d165731723e9d7eb068d8ed38bfac

Resource: https://developer.apple.com/videos/play/wwdc2024/10148

Kyle-Ye:
> Is it possible to unify the window title bar style for navigation and detail view? I think it would look better if they all have the style of the detail window

image

image

Do you mean such style? 

- By applying some List and Windows modifier
- Note some Windows modifier needs macOS 15+
- Also the Windows modifier only work if we go through SwiftUI.Window not the `StaticWindow` by AppKit

May not have time for a full PR on communication about the desired style.

Here is a PoC branch for SwiftUI Windows and List customization if someone is willing to pick and merge it.

https://github.com/tisfeng/Easydict/commit/5c8a22486a5b0a86b4585ced4f281bc19d263121

PR #739, opened 2024-11-30
Add SwiftUI window and list/window style

tisfeng:
From https://github.com/tisfeng/Easydict/pull/716#issuecomment-2508090759


AkaShark (dismissed):
LGTM

tisfeng:
I removed StaticWindows, and replaced the AppKit window with SwiftUI window management.

Referring to this https://github.com/tisfeng/Easydict/pull/716#issuecomment-2508090759, I tried to unify the style of the acknowledgment navigation window, although there are still some differences.

If anyone can further improve it, feel free to modify the code.

Kyle-Ye:
The `AcknowList` we are using now is a List under the hood, but it does not give a customizable Item. The only thing we can modify is to add a modifier related to List outside. (Currently we are adding a sidebar style to it.)

```swift
    var body: some View {
        NavigationStack {
            AcknowListSwiftUIView(acknowList: acknowList)
                .listStyle(.sidebar)
        }
        .hideWindowToolbarBackground()
        .thickMaterialWindowBackground()
    }
```

If this is enough for this project's customization, we can stick with it.

But if you want further customization, you can also give [my AcknowKit](https://github.com/Kyle-Ye/AcknowKit) a try. (Does not support auto discovery yet.)

Examples:

image

image

image

Easydict/Swift/View/AcknowListView/AcknowListView.swift
```diff
@@ -23,7 +23,10 @@ struct AcknowListView: View {
     var body: some View {
         NavigationStack {
             AcknowListSwiftUIView(acknowList: acknowList)
+                .listStyle(.sidebar)
         }
+        .hideWindowToolbarBackground()
```
Kyle-Ye, line 28: This will cause a UI conflict issue when we scroll the content.

So I guess we'd better not hide the toolbar background.

Instead use other ways (eg. containerBackground modifier) to achieve the consist effect of List and detail view.

Easydict/Swift/View/AcknowListView/AcknowListView.swift
```diff
@@ -23,7 +23,10 @@ struct AcknowListView: View {
     var body: some View {
         NavigationStack {
             AcknowListSwiftUIView(acknowList: acknowList)
+                .listStyle(.sidebar)
```
Kyle-Ye, line 26: Actually I think Form + formStyle(.grounded) would be more fitted in this scene.

But it turns out AcknowListSwiftUIView has hardcoded something. And the result is the header become one of the child element instead of a Form header.

image

Easydict/Swift/View/SettingView/Tabs/TabView/AboutTab.swift
```diff
@@ -26,6 +26,8 @@ struct SettingsAboutTab: View {
 struct AboutTab: View {
     // MARK: Internal
 
+    @Environment(\.openWindow) var openWindow
```
Kyle-Ye, line 29: Nit:

```suggestion
    @Environment(\.openWindow) private var openWindow
```

Kyle-Ye (dismissed):
LGTM for the StaticWindows change. And there is a UI issue I think we should fix before the merge. (Hide nav bar cause UI overlap when we scroll the content up in Acknow Window)

PR #744, opened 2024-12-08
fix: alert text show up to 1000

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/742

### DIFF
--- a/Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/ServiceConfigurationSecretSectionView.swift
+++ b/Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/ServiceConfigurationSecretSectionView.swift
@@ -88,7 +88,7 @@ struct ServiceConfigurationSecretSectionView<Content: View>: View {
                 viewModel.reset()
             }
         } message: {
-            Text(viewModel.errorMessage)
+            Text(viewModel.errorMessage.prefix(1000))
         }
         .onDisappear {
             viewModel.invalidate()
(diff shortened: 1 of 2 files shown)


tisfeng:
I tried adding ScrollView and limiting the number of frame and lines of Text, but nothing worked.

Jerry23011:
GPT says alert's message view doens't have native support for scroll view, a workaround would be to use `.sheet` to define a custom view. 

https://github.com/user-attachments/assets/ea8c3c90-dd37-4c55-98d0-2b4126801d86

tisfeng:
I see, because [alert](https://developer.apple.com/documentation/swiftui/view/alert(_:ispresented:presenting:actions:message:)-8584l) can only display text-based information 

> On iOS, tvOS, and watchOS, alerts only support controls with labels that are [Text](https://developer.apple.com/documentation/swiftui/text). Passing any other type of view results in the content being omitted.
>
> Only unstyled text is supported for the message.

It's true that a sheet can use a ScrollView, but the alert style is more appropriate here, and  I don't want to spend the effort to modify the sheet into the alert style, it's too tiring.

There are only a few cases where an alert would cause an exception, if there is no better solution, let's simply truncate the text.

Jerry23011:
> It's true that a sheet can use a ScrollView, but the alert style is more appropriate here, and I don't want to spend the effort to modify the sheet into the alert style, it's too tiring.
>
> There are only a few cases where an alert would cause an exception, if there is no better solution, let's simply truncate the text.

True true, let's keep it alert.

Jerry23011 (approved):

PR #755, opened 2024-12-22
fix: update Youdao web api secret key

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/754#issuecomment-2558422606


Jerry23011 (approved):
Works on my end. LGTM

PR #779, opened 2025-01-06
Fix acknow window

tisfeng:
https://github.com/tisfeng/Easydict/issues/767#issuecomment-2571669320

### DIFF
--- a/Easydict/App/EasydictApp.swift
+++ b/Easydict/App/EasydictApp.swift
@@ -76,13 +76,23 @@ struct EasydictApp: App {
                 .environment(\.locale, .init(identifier: I18nHelper.shared.localizeCode))
         }
 
+        // Can't use else direct here since SceneBuilder does not implement it yet.
+        // Fix issue "Closure containing control flow statement cannot be used with result builder 'SceneBuilder'"
+        if #available(macOS 15.0, *) {
+            // Acknowledgements window
+            Window("setting.about.acknowledgements", id: .acknowledgementsWindowId_15Plus) {
+                AcknowListView()
+            }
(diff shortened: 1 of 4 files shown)


Easydict/App/EasydictApp.swift
```diff
@@ -76,7 +76,20 @@ struct EasydictApp: App {
...
+            }
+            .windowToolbarStyle(.unifiedCompact)
+            .defaultSize(width: 800, height: 600)
+            .restorationBehavior(.disabled)
+        }
+
+        // Acknowledgements window for macOS 15-, cannot use #unavailable
         Window("setting.about.acknowledgements", id: .acknowledgementsWindowId) {
             AcknowListView()
         }
```
tisfeng, line 95: This code is not good. I tried to improve it, but didn't succeed. I know this is to pass the compilation. This is a problem with SwiftUI.

It seems we have no way to control a view to only display on macOS 15 or below, which will cause both Acknowledgements windows to be executed, including AcknowListView.

> rm -rf ~/Library/Saved\ Application\ State/com.izual.Easydict.savedState

If this solution works, I suggest we do not add this code.

Kyle-Ye: +1

Kyle-Ye: Can we request the original issue creator to try the solution first. The window should not be saved in most cases. (That's explained why you and I both can't reproduce it.)

Kyle-Ye (approved):

phlpsong (approved):

AkaShark (approved):

PR #807, opened 2025-02-05
fix: disable crash log is not working

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/805#issuecomment-2633515729

We should call `didSetAllowCrashLog()` when app launches to make sure enable or disable crash log.


AkaShark (approved):
LGTM

PR #809, opened 2025-02-05
fix: update Sentry to 8.44.0, to fix profiler cause memory leak

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/803#issuecomment-2635566465


AkaShark (approved):
LGTM

PR #815, opened 2025-02-10
feat: add an option to hide <think> tag content, hidden by default

tisfeng:
Close https://github.com/tisfeng/Easydict/issues/806

image

image


phlpsong (approved):
LGTM

PR #822, opened 2025-02-19
feat: add Built-in deepseek api support

garethng:
add DeepSeek Dictionary with Built-in api-endpoint and model name to avoid such issue like https://github.com/tisfeng/Easydict/issues/818

image
image


phlpsong:
Changed base to `dev`.

Easydict/App/Info-debug.plist
```diff
@@ -67,4 +67,4 @@
 	<key>SUEnableAutomaticChecks</key>
 	<true/>
 </dict>
-</plist>
+</plist>
```
phlpsong, line 70: The changes to these two plist files do not seem necessary, and it would be best removed in the commit.

garethng: The new line issue in Info.plist has been solved. The changes are removed now. Please help check with It. Thanks for your review.

Easydict/Swift/Service/DeepSeek/DeepSeekService.swift
```diff
@@ -0,0 +1,73 @@
...
+    // MARK: Lifecycle
+
+    override init() {
+        super.init()
+
+        // Set default supported models, disable user to change it.
+        // Generally, it should be updated only when the app is updated.
+        supportedModels = defaultModels.joined(separator: ", ")
+    }
+
```
tisfeng, line 25: Usually, we do not need to manually set `supportedModels` here, as this can prevent users from fixing the supported models.

Only certain special built-in services, such as `BuiltInAIService`, require this kind of initialization.

For the DeepSeek Service, we should allow users to modify it manually, as DeepSeek may later launch some new models, and Easydict may not be able to update in time.

tisfeng, line 25:
```suggestion

```

garethng: commit as reviewed. Thanks for your reviewing

tisfeng (approved):
Looks good.

Thanks your PR!

phlpsong (approved):

Jerry23011 (approved):

wjoyyy:
我这个是怎么回事呀？

tisfeng:
你的图片上传失败了。

PR #825, opened 2025-02-24
feat: add check for updates in menu

phlpsong:
About window size change:
Screenshot 2025-02-24 at 22 26 39

### DIFF
--- a/Easydict/Swift/View/MenuView/MainMenuCommand.swift
+++ b/Easydict/Swift/View/MenuView/MainMenuCommand.swift
@@ -30,6 +30,15 @@ struct EasydictMainMenu: Commands {
                 Text("menubar.about")
             }
         }
+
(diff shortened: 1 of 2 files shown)


Jerry23011:
Perhaps we could move the icon up a bit so that it is centered vertically
image

phlpsong:
@Jerry23011  I checked here, it looked like the logo image is in the center of `HStack`, not sure why top margin is larger.

Jerry23011:
> @Jerry23011 I checked here, it looked like the logo image is in the center of `HStack`, not sure why top margin is look bigger.

I believe it's the title bar
image

I just checked the current version and it's not centred either, so I say we're good with this.

Jerry23011 (approved):

AkaShark (approved):
LGTM

PR #877, opened 2025-04-19
feat: add GitHub Models service

tisfeng:
GitHub Models https://github.com/marketplace/models/azureml-meta/Llama-4-Maverick-17B-128E-Instruct-FP8/playground

image


tisfeng:
Since some reasoning models, like o4-mini only support temperature = 1, so I added an option for temperature, default to 0.3.

### temperature = 0.3 
image

### temperature = 1.0
image

Easydict/Swift/View/SettingView/Tabs/ServiceConfigurationView/SliderCell.swift
```diff
@@ -0,0 +1,70 @@
...
+    var body: some View {
+        HStack {
+            Slider(
+                value: $value,
+                in: minValue ... maxValue,
+                step: step
+            ) {
+                Text(titleKey)
+            } minimumValueLabel: {
+                Text(minValue.oneDecimalString)
```
phlpsong, line 58: We can also use `Text("\(number, specifier: "%.1f")")` here.

tisfeng: I know, since we need to use it in many place, so I added an Double extension, which can be used conveniently.

```swift
extension Double {
    var oneDecimalString: String {
        String(format: "%.1f", self)
    }
```

phlpsong (approved):

PR #910, opened 2025-06-02
Feat: add option for mini window position

tisfeng:
Fix https://github.com/tisfeng/Easydict/issues/839#issuecomment-2763040654

Just added a new option for mini window position, default is `mouse position`, which keeps the same showing position as before and has no impact on existing users.

image

### DIFF
--- a/Easydict/App/Localizable.xcstrings
+++ b/Easydict/App/Localizable.xcstrings
@@ -6470,6 +6470,34 @@
         }
       }
     },
+    "setting.advance.window.mini_window_position" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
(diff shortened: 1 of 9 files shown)


phlpsong:
I have an issue when debug app started, the main window will automatically move to the screen left corner, I tried dev branch which is normal. Could you double-check the behavior?

tisfeng:
ok, I will take a look later.

tisfeng:
> I have an issue when debug app started, the main window will automatically move to the screen left corner, I tried dev branch which is normal. Could you double-check the behavior?

Yes, your test is very careful, this is because I forgot to init `screenVisibleFrame`, I've fixed it in cd48d006

Please test it again.

PR #916, opened 2025-06-17
fix:  accent preference wasn't applied when playing translated English text

tisfeng:
Here's a fix for the issue where your chosen accent preference wasn't applied when playing translated English text.

Previously, when playing translated text in the results view, if the target language was English, the system would ignore your British/American accent preference selected in settings. This resulted in American English being played by default.

This update ensures that in EZWordResultView.m, when playing text translated into English, your stored accent setting ("us" or "uk") is read and correctly passed as an accent parameter to the audio playback service.

This resolves the problem you reported in GitHub issue  https://github.com/tisfeng/Easydict/issues/897#issuecomment-2980013629  allowing your British/American pronunciation choice to take effect when playing translated results.

### DIFF
--- a/Easydict/App/Localizable.xcstrings
+++ b/Easydict/App/Localizable.xcstrings
@@ -6470,6 +6470,34 @@
         }
       }
     },
+    "setting.advance.window.mini_window_position" : {
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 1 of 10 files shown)


tisfeng:
This PR is submitted by [Google jules  ](https://jules.google.com/) 

Just a test, but looks good 🙂

phlpsong (approved):

PR #917, opened 2025-06-17
refactor: remove ali web api

tisfeng:
Refactor: Remove Alibaba Translate Web API and related code

The web API for Alibaba's translation service has a 300-character translation limit, which is no longer practical.
This modification removes the logic for translating through the web API and standardizes the use of Alibaba's official translation API (key-based method).

Specific changes include:
- Removed the `requestByWeb` method and related token acquisition logic from `AliService.swift`.
- The `translate` method in `AliService.swift` now always calls `requestByAPI`.
- Updated the configuration options for Alibaba Translate in the settings interface, removing the API type selector.
- Deleted the `aliServiceApiTypeKey` configuration item from `Defaults.Keys+Extension.swift`.
- Cleaned up references to deleted files in the project file.


tisfeng:
image

phlpsong (approved):

PR #918, opened 2025-06-18
feat: Make result view title area clickable to expand/collapse

tisfeng:
Implements the feature request in issue #912 https://github.com/tisfeng/Easydict/issues/845 .

The entire title bar of a result entry in Easydict is now clickable to expand or collapse the content. Previously, only the small arrow icon could be used for this action.

Changes:
- Added an NSClickGestureRecognizer to the `topBarView` in `EZResultView.m`.
- The gesture recognizer's action handler replicates the logic from the existing `arrowButton`'s click block to ensure consistent behavior.
- The `arrowButton` remains functional as before.
- Both clicking the title area and clicking the arrow button will now toggle the visibility of the result content and update the arrow icon's state.


tisfeng:
image

PR #924, opened 2025-06-26
fix(ollama): enable sentence mode by default for auto-translation

Hephaest:
## Description

I recently switched to using the Ollama service and noticed a significant issue affecting the user experience. When I select text and trigger the translation shortcut (Option+D), the fixed translation window appears, but the translation doesn’t execute.

This behavior only occurs with the Ollama service — other remote APIs I’ve used, such as OpenAI and Gemini, work as expected.

## Root Cause

The problem was traced to the **sentence mode** setting, which was unexpectedly set to false by default when using Ollama. It’s unclear why this default differs from other services, but it appears to be unintentional or inconsistent with expected behavior.

## Demo

I’ve fixed the bug and tested the fix locally. The translation now triggers as expected when using Ollama, and everything is functioning smoothly. 🎉

https://github.com/user-attachments/assets/9f2f796b-f3d0-42a1-bb64-2a64a8d82de4

### DIFF
--- a/Easydict/Swift/Service/Ollama/OllamaService.swift
+++ b/Easydict/Swift/Service/Ollama/OllamaService.swift
@@ -51,7 +51,7 @@ class OllamaService: BaseOpenAIService {
     }
 
     override var isSentenceEnabledByDefault: Bool {
-        false
+        true
     }
 
     override var isDictionaryEnabledByDefault: Bool {


Hephaest:
Hi @tisfeng Please help review this PR, and feel free to push the commit for further fix. 🙌🏼 

BTW, I love this project, thanks for bring it to the open source world!

tisfeng:
Thanks your PR!

> The problem was traced to the sentence mode setting, which was unexpectedly set to false by default when using Ollama. It’s unclear why this default differs from other services, but it appears to be unintentional or inconsistent with expected behavior.

Because I previously tested and found that the Ollama model's explanation of words and sentences was not very good, so it was disabled by default.

Maybe it has improved a lot now, and it's ok to enable it by default.

This way, it can also be consistent with OpenAI and Gemini, no problem.

Easydict/Swift/Service/Ollama/OllamaService.swift
```diff
@@ -50,14 +50,6 @@ class OllamaService: BaseOpenAIService {
...
     }
 
-    override var isSentenceEnabledByDefault: Bool {
-        false
-    }
-
-    override var isDictionaryEnabledByDefault: Bool {
-        false
-    }
-
```
tisfeng, line 60: I removed these override, since they are returned `true` in superclass.

Hephaest:
@tisfeng Thanks for reply, you actually helped me realize how I fell into this issue. Sadly, my original "fix" was totally wrong 😅 The real issue was that I had **turned off the translation toggle**. With all three toggles (translation, sentence, dictionary) off, the system didn’t trigger auto-translation — which makes sense now.


#### Why I Turn It Off?
I originally disabled the toggles because I thought they only applied to the built-in prompt. Since I was using a custom prompt, I assumed they weren’t needed. But now I realize in order to achieve auto query, I at least need to toggle **translation** on.

image

#### Fun Fact Asides:
The demo "worked" because I had sentence mode turned on by default – even with translation and dictionary toggles off, auto-translation was still triggered. I even tested with only the dictionary toggle on… and it works! 😂

Back to this PR, your fix is definitely more robust and aligns better with how other online models behave. With Ollama now supporting [non-Meta models](https://ollama.com/search), it makes sense to bring this into consistency across the board.

Thanks again — really appreciate the discussion and your help in figuring this out! 🙏

tisfeng:
ok, good 😊

tisfeng (approved):

PR #931, opened 2025-07-08
docs: fix a picture not show and add explanation

tonngw:
This PR modifies two documents file: `docs/How-to-use-macOS-system-dictionary-in-Easydict-zh.md` and `docs/How-to-use-macOS-system-dictionary-in-Easydict-en.md`.

Change:
1. Fixed the issue where the image "Open Dictionary Folder" was not displaying. It was caused by the image src line break.
image

document link at https://github.com/tisfeng/Easydict/blob/main/docs/How-to-use-macOS-system-dictionary-in-Easydict-zh.md
it can now be displayed normally after fix the issue.
image

2. Added explanations about the dictionary folder and instructions on how to adjust the order of dictionaries.

The above changes from some confusion I encountered while following the documentation. I hope this improvement to the documents will be helpful.


tisfeng:
ok, thanks your PR

Issue #816, opened 2025-02-10
🚀 Feature Request: Keyboard Shortcut for "Replace with Translation"


victorcorreal:
### Please confirm the following:

- [x] I have carefully read the [README](https://github.com/tisfeng/Easydict#readme)
- [x] I have searched through the [issues](https://github.com/tisfeng/Easydict/issues) page and didn't find a similar feature request.(Including issue that has been closed)
- [x] Easydict has been upgraded to the [latest version](https://github.com/tisfeng/Easydict/releases)
- [x] I understand and agree to the above, and understand that the project maintainer has limited energy, **issues that do not follow the rules may be ignored or closed directly**

### Feature description

I'm trying to use the "Replace with Translation" feature. Is there a keyboard shortcut available for this function? If not, is it possible to configure one? I couldn't find relevant information in the documentation.

Thank you for your assistance.

### Use case

For example, the user selects a portion of text in the email draft in gmail that they wish to translate and replace with the translated version.

### Are you willing to submit a PR to implement this feature

- [ ] I'm willing to submit PR to implement this feature

tisfeng:
Currently, there is no shortcut key for this, as there may be multiple service translation results, and it cannot be determined which service to use for the translation replacement.

victorcorreal:
> Currently, there is no shortcut key for this, as there may be multiple service translation results, and it cannot be determined which service to use for the translation replacement.

In my case, I only use one; I only have one translation service activated. In fact, I find it hard to understand the usefulness of having more than one service to do the same thing. What I have done is try for a while, see which one is the best (or the one that best fits my needs), and keep only that one activated.

Having the option to select a text and automatically replace it with the translation without going through the widget, as other solutions do, would be wonderful. Let's see if someone who thinks the same appears.

Thank you.

tisfeng:
> Having the option to select a text and automatically replace it with the translation without going through the widget, as other solutions do, would be wonderful.

ok, I understand your thoughts.

This is a good idea, and we will consider it.

tisfeng:
[2.16.0](https://github.com/tisfeng/Easydict/releases/tag/2.16.0) version has implemented the "translate and replace" feature, which can be triggered by a shortcut key.

Currently, the built-in AI translation service is used by default, and it will later allow the selection of a translation service.

PR #954, opened 2025-08-31
feat(shortcuts): add 'Translate and Replace' and 'Polish and Replace' actions

tisfeng:
This PR primarily adds two shortcut actions, 'Translate and Replace' and 'Polish and Replace', for translating/polishing the selected text in an input box and then replacing it.

In addition to being triggered via application menu items, you can also assign global shortcuts to them for convenient access.

To avoid conflicts from Easydict having too many global shortcuts, I have not set default shortcuts for them.

By default, translation uses `BuiltInAIService`, and polishing uses `PolishingService`. Currently, these default services cannot be changed.

Close https://github.com/tisfeng/Easydict/issues/724 , https://github.com/tisfeng/Easydict/issues/922 , https://github.com/tisfeng/Easydict/issues/816

https://github.com/user-attachments/assets/ba8e7411-3782-4e33-b028-792d94732729

### DIFF
--- a/Easydict/Swift/Feature/ActionManager/ActionManager.swift
+++ b/Easydict/Swift/Feature/ActionManager/ActionManager.swift
@@ -182,8 +182,6 @@ class ActionManager: NSObject {
         _ contentStream: AsyncThrowingStream<String, Error>,
         textFieldInfo: TextFieldInfo
     ) async {
-        logInfo("Starting streaming text replacement with: \(textFieldInfo)")
-
         if textFieldInfo.isSupportedAXElement {
             // Streaming replacement for supported text fields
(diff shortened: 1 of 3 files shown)


tisfeng:
I've added an "Auto Select All" option, which is enabled by default.

For some applications that do not support AX text selection, this option is necessary. Otherwise, we cannot retrieve the currently selected text to perform translation or polishing.

Additionally, if an application does not support AX features, text replacement is achieved by simulating copy and paste. Similar to forced text capture, this may overwrite the contents of user system clipboard.

image

tisfeng:
Next, we can allow users to configure the default services for translation and polishing.

Issue #952, opened 2025-08-25
When will it be updated? icons for macos 26 tahoe


naixzh2024:
### 请先确认以下事项

- [x] 已仔细阅读了 [README](https://github.com/tisfeng/Easydict#readme)
- [x] 在 [issues](https://github.com/tisfeng/Easydict/issues) 页面搜索过（包括已关闭的 issue），未发现类似功能建议
- [x] Easydict 已升级到 [最新版本](https://github.com/tisfeng/Easydict/releases)
- [x] 我理解并认可上述内容，并理解项目维护者精力有限，**不遵循规则的 issue 可能会被无视或直接关闭**

### 功能描述

<img src="https://startupnews.fyi/wp-content/uploads/2025/07/messages-macos-26.jpg">

### 使用场景

mac

### 是否愿意提交 PR 实现该功能

- [ ] 我愿意提交 PR 实现该功能

tisfeng:
We plan to release a new version this weekend.

> icons for macos 26 tahoe

Do you mean the new Easydict logo?

Thanks for the reminder — we do need a new logo for macOS 26 Tahoe. 

But that's uncertain, because  I am not able to make a new logo; we may need to ask other teammates for help.

<img width="640" height="546" alt="Image" src="https://github.com/user-attachments/assets/2c3a633f-7793-4548-8470-f829370c4c48" />

tisfeng:
@Keldos-Li  Hello, we're planning to update the Easydict logo for macOS 26 Tahoe. Do you have time to help?

Keldos-Li:
> Do you have time to help?

Sure! I'm ready to use _Icon Composer_. But maybe you could wait for a few days? (I've been rushing to finish a report these days 🥲)

tisfeng:
ok, it isn't urgent, you can arrange the time yourself. 

Thank you!

Keldos-Li:
I submitted the icon update in #956 😃

tisfeng:
The new icon needs to be packaged and released using the latest Xcode 26 to be compatible with macOS 26.

However, when packaging and releasing version 2.15.0 using Xcode 26, serious memory leaks and issues with the input cursor were encountered: https://github.com/tisfeng/Easydict/issues/970

We haven't had time to fix the above two issues yet, so the 2.16.0 version will temporarily use the old icon.

KirkifyAIGenerator:
thx

Keldos-Li:
[...]

PR #956, opened 2025-09-03
per(UI): update to new icon design using Icon Composer

Keldos-Li:
## Note
- Based on the original design, all drawing effects were removed and replaced with App Composer for automatic rendering;
- Dark mode uses inverted colors;
- Colors, sizes, and positions were slightly adjusted;
- I'm not quite sure if the icons will look blurry when exported at smaller sizes. It feels like applying the glass effect to the bubble border and the character "文" adds a bit too much detail. However, I haven't updated to macOS 26, so I'm uncertain about the actual effect. I'm also not sure if this icon is suitable for non-macOS26 users.

Feel free to suggest modifications~

Close https://github.com/tisfeng/Easydict/issues/952 ,  https://github.com/tisfeng/Easydict/issues/972

## Preview
image
image
image
image
image


tisfeng:
Ok, it looks great at first glance. I'll run it later to see the actual effect.

Please share your thoughts on the effect of the new logo on macOS 26 and earlier systems.

tisfeng:
The new logo looks good on macOS 26, it's adapted to the new system.

image

image

tisfeng:
> I'm not quite sure if the icons will look blurry when exported at smaller sizes.

There is a small issue, compared to the old logo, the new logo looks a bit blurry.

Is this an intentional change?

I feel like it would be better if it were clearer.

image

tisfeng:
Please share the effect of the new logo on the old system, such as macOS 15.

Jerry23011:
> Please share the effect of the new logo on the old system, such as macOS 15.

Not sure what happened here. macOS 15 shows a blank icone.

It could be due to using Pod to configure Easydict's icon, and Xcode doesn't recognizing the new icon.

Letting Easydict using Xcode's default icon configuration may help - macOS should fall back on using `Assets.xcassets` with macOS 15.
Screenshot 2025-09-06 at 04 37 52

Keldos-Li:
> There is a small issue, compared to the old logo, the new logo looks a bit blurry.
>
> Is this an intentional change?

Yes, I also feel it's a bit blurry, but it’s not intentional \_(:з」∠)\_. I think after Apple used the liquid glass effect to present icon elements, many icon details combined with the highlight reflections do become automatically a bit blurry, with less clear edges, and the rendering resolution isn’t as high as the previous manual control.

However, I did reduce the opacity, so the black is now somewhat transparent, not the pure black as before (because it needs to achieve the glass effect). This might also be a reason for the slight blurriness.

Perhaps I can make the icon background color whiter in the light mode; maybe that would improve it a bit.

tisfeng:
ok, new icon looks better

image

tisfeng:
> > Please share the effect of the new logo on the old system, such as macOS 15.
> 
> Not sure what happened here. macOS 15 shows a blank icone.
> 
> It could be due to using Pod to configure Easydict's icon, and Xcode doesn't recognizing the new icon.
> 
> Letting Easydict using Xcode's default icon configuration may help - macOS should fall back on using `Assets.xcassets` with macOS 15. Screenshot 2025-09-06 at 04 37 52

so strange, let's look into it later

Keldos-Li:
The app I exported using `Product > Archive` doesn't seem to have any icon missing issues on my Mac. Could it be a problem with the debug build's separate icon setting?

image
image

Jerry23011:
> The app I exported using `Product > Archive` doesn't seem to have any icon missing issues on my Mac. Could it be a problem with the debug build's separate icon setting?
> 
> image image

Would it be on Xcode 26 beta? I complied on macOS 15 with Xcode 15

Keldos-Li:
> Would it be on Xcode 26 beta? I complied on macOS 15 with Xcode 15

Yes, the new icon file format is different, and you must use the new Xcode-beta to get the rendered icon correctly.

tisfeng:
> The app I exported using `Product > Archive` doesn't seem to have any icon missing issues on my Mac. Could it be a problem with the debug build's separate icon setting?
> 
> image image

@Jerry23011 

Have you tested with this method yet, and are there still any issues?

I plan to release a new version in the next couple of days, and I want to merge this PR.

Jerry23011:
> @Jerry23011
> 
> Have you tested with this method yet, and are there still any issues?
> 
> I plan to release a new version in the next couple of days, and I want to merge this PR.

Yes, I'm on macOS 15.7 with Xcode 26.0. Archive works.

Thank you for pointing this out @Keldos-Li 

Screenshot 2025-09-17 at 20 05 16

Jerry23011 (approved):
LGTM

Thank you for the new icon! @Keldos-Li

PR #960, opened 2025-09-06
fix(eudic): ensure Eudic quick link icon is always shown when enabled

tisfeng:
Removes the unreliable installation check for the Eudic application.

Previously, the quick link icon would not appear for users with certain versions of Eudic due to an incomplete list of bundle IDs. This change fixes that issue by removing the check entirely. The icon will now always be displayed if the user enables it in the settings, leaving the responsibility of having the app installed to the user.

The description for the corresponding setting has also been updated to remove the "(if installed)" text.

Fix https://github.com/tisfeng/Easydict/issues/957#issuecomment-3261505123

### DIFF
--- a/Easydict/App/Localizable.xcstrings
+++ b/Easydict/App/Localizable.xcstrings
@@ -8192,25 +8192,25 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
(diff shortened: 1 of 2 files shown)


Jerry23011 (approved):
LGTM

PR #976, opened 2025-09-17
feat(selection): prioritize AppleScript for browser text selection

tisfeng:
The Accessibility API can be unreliable for fetching selected text in some web browsers. To provide a more robust experience, this commit introduces an option to prioritize AppleScript for this task.

A new setting, `preferAppleScriptAPI`, has been added. When enabled, the application will attempt to get the selected text from supported browsers (e.g., Chrome, Safari) using AppleScript first. If this method fails, it gracefully falls back to the existing Accessibility API-based approach, ensuring text is still captured.

Fix https://github.com/tisfeng/Easydict/issues/944

### DIFF
--- a/Easydict/Swift/Feature/Configuration/Configuration.swift
+++ b/Easydict/Swift/Feature/Configuration/Configuration.swift
@@ -29,6 +29,9 @@ enum EnglishPronunciation: Int {
 
 // MARK: - Configuration
 
+/// Singleton class to manage application configuration settings.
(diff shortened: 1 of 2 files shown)


Jerry23011 (approved):
LGTM

I tested science.org and it works with/without prioritize AppleScript enabled. However, I would say it wouldn't affect existing user experience.

It would be great if anyone who can reproduce the issue in #944 can test themselves on that specific use case

tisfeng:
ok, lets' merge it.

Issue #982, opened 2025-09-18
🐞 Bug Report: default keyboard commands conflict with many other applications


grantbarrett:
### Please confirm the following:

- [x] Be sure to check out [FAQ](https://github.com/tisfeng/Easydict/wiki/FAQ)
- [x] I have carefully read the [README](https://github.com/tisfeng/Easydict#readme)
- [x] I have searched through the [issues](https://github.com/tisfeng/Easydict/issues) page but couldn't find a solution.(Including issue that has been closed)
- [x] Easydict has been upgraded to the [latest version](https://github.com/tisfeng/Easydict/releases)
- [x] I understand and agree to the above, and understand that the project maintainer has limited energy, **issues that do not follow the rules may be ignored or closed directly**

### Problem description

I just updated to 2.16.0. The default keyboard commands now conflict system-wide with many other applications. For example, ⌘T is "new tab" in many browsers and other applications, but because Easydict captures the keyboard command everywhere on the system, not just while you are in the Easydict window, ⌘T no longer functions elsewhere. Similarly, many of the defaults, such as ⌘S, are decades-old commands (this one for "Save") that should never be set as defaults. In other words, the App Shortcut keyboard commands are behaving like Global Shortcut keyboard commands, when they should not.

### Is the issue consistently reproducible?

Reproducible

### Steps to Reproduce

To reproduce, try one of the default commands in Easydict 2.16.0, but in another applications. 

### Screenshot

<img width="729" height="635" alt="Image" src="https://github.com/user-attachments/assets/3010f3f4-af05-471b-9583-5d287be6d4a9" />

### Runtime log(zip file)

[Easydict log 2025-09-18 16-07-19-752.zip](https://github.com/user-attachments/files/22415957/Easydict.log.2025-09-18.16-07-19-752.zip)

### Expected result

The result is that App Shortcuts should not interfere outside of the Easydict windows.

### Proposed solution (optional)

A temporary workaround is to disable the interfering default keyboard commands by clicking the X to the right of each. 

A permanent solution would be to make sure the "App Shortcut" default keyboard commands only work when the Easydict window is in the foreground. 

### Device type and OS version

[...]

PR #999, opened 2025-09-28
fix(shortcut): prevent in-app shortcuts from being registered globally

tisfeng:
Previously, some shortcuts intended for in-app use were incorrectly registered as system-wide hotkeys. This could lead to conflicts with other applications or system functions.

This commit refactors the shortcut management logic to clearly differentiate between global and in-app shortcuts. An explicit `globalActions` list has been introduced to define which actions should have global hotkeys. The binding function now verifies an action against this list before registering it with the `HotKeyCenter`, ensuring only designated shortcuts are active system-wide.

Bug introduced in https://github.com/tisfeng/Easydict/pull/953

Fix https://github.com/tisfeng/Easydict/issues/997 , https://github.com/tisfeng/Easydict/issues/992 , https://github.com/tisfeng/Easydict/issues/982

### DIFF
--- a/Easydict/Swift/Feature/Shortcut/Model/ShortcutManager+Default.swift
+++ b/Easydict/Swift/Feature/Shortcut/Model/ShortcutManager+Default.swift
@@ -23,7 +23,9 @@ extension ShortcutManager {
         Defaults[.snipShortcut] = KeyCombo(key: .s, cocoaModifiers: .option)
         Defaults[.selectionShortcut] = KeyCombo(key: .d, cocoaModifiers: .option)
         Defaults[.showMiniWindowShortcut] = KeyCombo(key: .f, cocoaModifiers: .option)
-        Defaults[.silentScreenshotOCRShortcut] = KeyCombo(key: .s, cocoaModifiers: [.option, .shift])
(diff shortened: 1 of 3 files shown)


Easydict/Swift/Feature/Shortcut/Model/ShortcutManager+Default.swift
```diff
@@ -50,27 +52,20 @@ extension ShortcutManager {
...
 
-    /// Bind default shortcut action
-    func bindingShortcutAction(keyCombo: KeyCombo?, action: ShortcutAction) {
+    /// Bind global shortcut action (registers as system-wide hotkey)
+    func bindingGlobalShortcutAction(keyCombo: KeyCombo?, action: ShortcutAction) {
         HotKeyCenter.shared.unregisterHotKey(with: action.rawValue)
-        guard let keyCombo else {
+
+        // Ensure the action is a global action and keyCombo is valid
+        guard let keyCombo, globalActions.contains(action) else {
```
tisfeng, line 68: This is the key code to fix the bug.

AkaShark (approved):

PR #1004, opened 2025-10-03
fix(ocr): handle empty language probabilities for numeric input

tisfeng:
When performing OCR on text containing only numbers, the AppleLanguageDetector returns an empty `rawProbabilities` dictionary. This unhandled case caused the AppleOCREngine to fail, resulting in an empty recognition result.

This patch addresses the issue by updating the `smartMerging` condition in the OCR engine to be true if `rawProbabilities` is empty. This ensures that purely numeric text is processed correctly.

Additionally, a comment has been added to `AppleLanguageDetector` to clarify that `rawProbabilities` can be empty in edge cases like this.

Closes: #1001


AkaShark (approved):
LGTM

PR #1008, opened 2025-10-06
Fix tencent translate

tisfeng:
Try to fix https://github.com/tisfeng/Easydict/issues/975


tisfeng:
Referring to the reply in https://github.com/Alamofire/Alamofire/issues/3983 , I fixed the fundamental issue that upgrading Alamofire would cause the Tencent translation to fail.

Everyone please note that when we use Alamofire later, parameters should use the `Parameters` type, not `[String: Any]`.

PR #846, opened 2025-03-21
refactor: rewrite screenshot feature with Swift

tisfeng:
Close https://github.com/tisfeng/Easydict/issues/22 , https://github.com/tisfeng/Easydict/issues/595 , https://github.com/tisfeng/Easydict/issues/736 , https://github.com/tisfeng/Easydict/issues/461

https://github.com/tisfeng/Easydict/issues/81

Support clicking D to select the last screenshot area.

PR #1020, opened 2025-10-19
refactor(screenshot): replace legacy Objective-C snip with new Swift implementation

tisfeng:
This commit refactors the screenshot functionality by completely removing the old Objective-C-based 'Snip' library. It is replaced by the modern, pure Swift 'Screenshot' implementation.

Key changes include:
- Deleted all files related to the 'Snip' library.
- Updated EZWindowManager to use the new 'Screenshot' class for capturing screen areas.
- Removed the beta feature flag that previously allowed switching between the old and new screenshot implementations.
- Removed the 'Snip' library from the acknowledgements list.
- Cleaned up the Xcode project file to remove references to the deleted files.

Close https://github.com/tisfeng/Easydict/issues/983


tisfeng:
New Swift screenshot is tested by time https://github.com/tisfeng/Easydict/pull/846 , we can use it in release version for all users, so removed the old objc snip code.

PR #1049, opened 2026-01-10
Add section to tell if it's free or pro based on apikey

purifier1990:
Screenshot 2026-01-10 at 18 47 46

Also added i18n strings

Close https://github.com/tisfeng/Easydict/issues/932


Jerry23011:
Thank you for the PR, and I love the idea to make it clearer for users.

I went over the issue mentioned above. I think 沉浸式翻译 has a great naming: 免费/高级 | Free/Advanced

Since we're an open-source software, perhaps it's better not to tier users. 

I have several ideas here:
- 内置/高级 | Built-in/Advanced
- 免费/高级 | Free/Advanced （沉浸式翻译）
- 开箱即用/需要密钥 | Built-in/Requires Setup (could be a bit long, but my personal favourite)

Nevertheless, I think Free/Pro already works well, feel free to consider the above.

purifier1990:
> Thank you for the PR, and I love the idea to make it clearer for users.
> 
> I went over the issue mentioned above. I think 沉浸式翻译 has a great naming: 免费/高级 | Free/Advanced
> 
> Since we're an open-source software, perhaps it's better not to tier users.
> 
> I have several ideas here:
> 
> * 内置/高级 | Built-in/Advanced
> * 免费/高级 | Free/Advanced （沉浸式翻译）
> * 开箱即用/需要密钥 | Built-in/Requires Setup (could be a bit long, but my personal favourite)
> 
> Nevertheless, I think Free/Pro already works well, feel free to consider the above.

To me, I will prefer following trending tools' wording, the Builtin/Requres setup seems not have the same patterns, I will vote for Free/Advanced, if Free/Pro is not a good candidate, welcome to vote for others

Jerry23011 (dismissed):
~~LGTM. @purifier1990 Feel free to push the Free/Advanced changes. Thank you for your PR!~~

Jerry23011 (commented):
After taking a closer look, users can use DeepL and several others without an API key, so I'm thinking how we can let users know about that? Or do we also want to move them to Free?

purifier1990:
> After taking a closer look, users can use DeepL and several others without an API key, so I'm thinking how we can let users know about that? Or do we also want to move them to Free?

@Jerry23011 I think I was mainly depended on what the code already has, like builtin enum and the StreamService, I thought this kind of StreamService should be based on some keys, like what AI did for this code, if this looks not what you expect, maybe the type is being abused using too much?

tisfeng:
Thanks for your PR, I generally agree with the Free/Pro classification, and I'll make some optimizations and simplifications to it later.

tisfeng:
Later I will continue to simplify the code by removing things like API keys for built-in traditional translation services(like tencent api key) and code such as `hasFreeQuotaLeft`, which I think are no longer needed and just add complexity to the code.

tisfeng:
please review

PR #1059, opened 2026-01-16
fix(event-monitor): resolve click query interception issue

tisfeng:
This commit fixes the issue where click queries were being intercepted prematurely by the local event listener's leftMouseDown handler.

The problem was that the leftMouseDown event was always triggering the pop button dismissal logic before the click action could be processed, preventing valid click queries from working.

The fix adds a check to skip dismissal when the mouse click occurs within the pop button window itself, allowing the pop button's click action to fire properly before any dismissal logic runs.

This ensures that click queries work correctly while maintaining the pop button functionality.

Fix https://github.com/tisfeng/Easydict/issues/1053


tisfeng:
please review

PR #1060, opened 2026-01-16
fix(event-monitor): remove local monitor to restore pop button click in issue #1057

xyzhou-1:
Revert to global-only monitoring like 2.16.2. 
https://github.com/tisfeng/Easydict/blob/821e7d7334b9937eaaf79cd8ee34f537677d0b21/Easydict/objc/EventMonitor/EZEventMonitor.m#L181-L189
The Swift EventMonitor used local+global, which intercepted clicks early and dismissed the pop button before the click handler ran. This rollback prevents premature dismiss and allows the translation window to open.

Also remove some unused code.

### DIFF
--- a/Easydict/Swift/Feature/Configuration/MyConfiguration.swift
+++ b/Easydict/Swift/Feature/Configuration/MyConfiguration.swift
@@ -415,7 +415,7 @@ extension MyConfiguration {
     }
 
     fileprivate func didSetAutoSelectText() {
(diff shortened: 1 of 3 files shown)


tisfeng:
Thanks your PR, I has fixed this issue by https://github.com/tisfeng/Easydict/pull/1059 

It's not good to remove local monitor.

PR #1070, opened 2026-01-25
fix: remove macOS Icon file that breaks Windows compatibility

xiaocang:
The `Icon\r` file is a macOS Finder custom icon file that contains a carriage return character in its filename. This character is invalid on Windows, causing `git clone` and `git reset` to fail with:

> error: invalid path 'Icon?'


tisfeng:
Thanks your PR, looks good

PR #1071, opened 2026-01-26
Add 10% 20%  30%  and 40% options for max window height percentage

yhzhang067:
This PR adds four new options (10% 20% 30% and 40%) for the max window height percentage configuration, allowing users to set smaller window heights if needed.


tisfeng:
Thanks your PR, looks good

PR #1084, opened 2026-02-02
feat(selection): add conditional triggers for auto show query icon

tisfeng:
Add configurable conditions to control when the query icon appears after text selection:

- Language-based filtering: only show when selected text language differs from excluded language
- Minimum character length requirement: customizable threshold (0-50 characters) to reduce accidental triggers
- Settings UI in Advanced tab to configure both conditions

The query icon now respects these conditions to provide more precise control over auto-selection behavior.
image


tisfeng:
closed by https://github.com/tisfeng/Easydict/pull/1085

PR #1092, opened 2026-02-04
fix(pdf): fix query icon not showing in PDF text selection

tisfeng:
This commit fixes the issue where the query icon was not appearing when selecting text in PDF documents.

The root cause was that PDF content uses the AXGroup role, which was not included in the selectable text roles. Adding kAXGroupRole to the selectableTextRoles set enables the query icon to appear correctly in PDF documents.

Fix https://github.com/tisfeng/Easydict/issues/1091

### DIFF
--- a/Easydict/App/Localizable.xcstrings
+++ b/Easydict/App/Localizable.xcstrings
@@ -8357,25 +8357,25 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Free"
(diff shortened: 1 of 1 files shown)


Jerry23011:
@linxinyaooo Feel free to test out this PR. Thanks

tisfeng:
> Did not test myself as I don't use the software, code itself LGTM

You don't need to use any other software; just open a PDF file, and then you can test this issue.

PR #1138, opened 2026-04-03
Add es_ES locale

terorero:
Add es_ES locale (verified)


MoonMao42 (commented):
Thanks for adding Spanish — the project only has en/sk/zh-Hans/zh-Hant right now.

That said, the translation needs some work before it's ready to merge. I went through the `Localizable.xcstrings` diff and found a few issues:

**Partially translated entries** — Several strings have only the first word or verb translated while the rest stays in English. For example:

- `"Limpiar input when translating"` (should be fully Spanish)
- `"Ocultar Menu Bar Icon"`
- `"Abrir in Apple Dictionary"`
- `"Traducir the following ${{queryFromLanguage}} text into ${{queryTargetLanguage}}: ${{queryText}}"`
- `"Habilitar Force Get Selected Text"`
- `"Habilitar OCR Text Format Optimization"`

There are about 10+ entries like this. They look like an interrupted pass — the first verb got translated but the sentence wasn't finished.

**Formatting changes** — The diff is ~15k lines but a big chunk is just whitespace reformatting (` "value" : ` → `"value":`) across existing en/sk/zh entries. This makes the diff hard to review and could cause merge conflicts. Would be better to keep only the actual `es` additions.

**58 entries identical to English** — Some are fine (brand names like "DeepL", "Baidu"), but others like `"Buffer Too Small"` or `"Decode Error"` should probably be translated.

Could you do another pass on the incomplete translations.

terorero:
Please review "The Fix the incomplete translations"

tisfeng:
Please resolve conflicts and push it forward

PR #1151, opened 2026-04-18
fix(settings): remove window resize animation lag on tab switch

### LINKED ISSUE #1149: 🐞 反馈问题：设置界面卡顿
### 请先确认以下事项：

- [x] 请务必查看 [常见问题](https://github.com/tisfeng/Easydict/wiki/%E5%B8%B8%E8%A7%81%E9%97%AE%E9%A2%98)
- [x] 已仔细阅读了 [README](https://github.com/tisfeng/Easydict#readme)
- [x] 在 [issues](https://github.com/tisfeng/Easydict/issues) 页面搜索过问题（包括已关闭的 issue），但未能找到解决方法
- [x] Easydict 已升级到 [最新版本](https://github.com/tisfeng/Easydict/releases)
- [x] 我理解并认可上述内容，并理解项目维护者精力有限，**不遵循规则的 issue 可能会被无视或直接关闭**

### 问题描述

设置界面很卡, 如视频所示, 在"通用"和"关于"这两个 tab 切换时, 点击后需要 2-3 秒才有反应

### 该问题是否可以稳定重现？

可重现

### 重现步骤

在"通用"和"关于"这两个 tab 切换时, 点击后需要 2-3 秒才有反应

### 截图

https://github.com/user-attachments/assets/08d35002-f121-49f2-9004-3ad5f132fbca

### 运行日志（zip 文件）

[Easydict log 2026-04-17 20-04-11-896.zip](https://github.com/user-attachments/files/26824043/Easydict.log.2026-04-17.20-04-11-896.zip)

### 期望结果

不卡

### 解决方案（可选）

_No response_

### 设备型号 && 系统版本

Apple M4 Pro 26.4.1 (25E253)

### 是否愿意提交 PR 解决该问题？

- [ ] 我愿意提交 PR 来解决该问题

MoonMao42:
## Summary

Fix #1149.

Switching between "General" and "About" in the settings window took 2-3s to respond, with the window visibly shrinking or growing before another click registered.

`SettingView.resizeWindowFrame()` animates the window to the new tab height, then uses `DispatchQueue.main.asyncAfter(0.3)` to remove `.resizable` from the style mask after the animation "should" be done. The 420pt height delta between General (720pt) and About (300pt) pushes `NSWindow.animationResizeTime` near its upper bound (~1.2s), and the 0.3s guard stacks on top. Other tab pairs share 720pt so their delta is zero and the problem never surfaces there.

## Changes

- Flip `animate: true` → `animate: false` on `window.setFrame`
- Remove `.resizable` synchronously; drop the `asyncAfter(0.3)` timer

## Notes

This PR only addresses the lag caused by the window resize animation. Settings tabs with more content (such as General and Advanced) may have separate SwiftUI rendering costs, which can be tracked separately in a follow-up.​​​​​​​​​​​​​​​​


tisfeng:
it's ok remove the animation, but the delay in switching tabs here may be related to the data loading of the tab view. The general and advanced Tab pages have too much data, causing stuttering.

It needs to be solved in this direction.

MoonMao42:
Thanks for the review. I want to share some measurements on the data loading angle: the symptoms don't line up with that hypothesis.

**Observed behavior**:

| Tab pair | Height delta | Behavior (both directions) |
|---|---|---|
| General ↔ About | 420pt | Lags ~2-3s |
| General ↔ Advanced | 0pt | No lag |
| General ↔ Service | 0pt | No lag |

**What removing the animation actually changes**:

- **Data loading itself**: no effect. It takes exactly as long as it did before.
- **User-visible behavior**: the `.task { fetchRepoLatestVersion(...) }` in GeneralTab is async and doesn't block the main thread. With or without the animation, it updates a text label when the result arrives. No blank state, no stutter.
- **What goes away**: the 1.2s frame animation, and the ~2-3s main-thread freeze that came with it.

tisfeng:
I agree with removing the animation, there's no problem with that.

What I want to say is that when I tested it on my computer, there was a noticeable delay when switching between Tabs with a lot of data content like GeneralTab and AdvancedTab. I suspect it might be related to reading user preference settings data, as there's too much data to read, causing UI rendering to take a lot of time.

However, for PrivacyTab and AboutTab, which have very little user data, switching Tabs is very smooth.

I've added a test branch [test/settings-tab-lag](https://github.com/tisfeng/Easydict/tree/test/settings-tab-lag). After replacing the GeneralTab body content with a simple image, the GeneralTab Tab switching no longer has a delay.

Please see the video for details.

https://github.com/user-attachments/assets/cef8b826-bd8f-4191-bb20-d33d732c9c89

tisfeng:
This issue may have been around for a long time and it's ok to put it in another PR.

MoonMao42:
> This issue may have been around for a long time and it's ok to put it in another PR.

 I checked your branch and it actually includes the commit from my current PR, then replaces GeneralTab with an image. That means the variables were not properly isolated.

  I created a branch called Easydict-real-test-settings-tab-la that removes my PR commit and only changes GeneralTab to an image. I tested that version and it did not solve the issue.

  So at this point, I’m more inclined to think the problem is caused by the window resize animation rather than by the amount of data in the tab.
  If you think `GeneralTab` / `AdvancedTab` still need refactoring or other changes to reduce rendering cost, I’d be happy to discuss possible directions in a separate PR.

tisfeng:
I agree this PR can stay focused on removing the window resize animation, and the code change itself looks fine.

But I don't think this fully rules out the rendering cost of GeneralTab / AdvancedTab. My point is not that we must solve that in this PR, but that the settings tabs with many controls and Defaults bindings still show noticeable switching cost on my machine, while lightweight tabs are smooth.

So please avoid framing the root cause as only the window animation or saying the data/rendering hypothesis does not line up. This PR can say it removes one visible source of lag from animated resizing. The heavier SwiftUI settings tab rendering can be tracked separately.

tisfeng (approved):

MoonMao42:
> I agree this PR can stay focused on removing the window resize animation, and the code change itself looks fine.
> 
> 
> 
> But I don't think this fully rules out the rendering cost of GeneralTab / AdvancedTab. My point is not that we must solve that in this PR, but that the settings tabs with many controls and Defaults bindings still show noticeable switching cost on my machine, while lightweight tabs are smooth.
> 
> 
> 
> So please avoid framing the root cause as only the window animation or saying the data/rendering hypothesis does not line up. This PR can say it removes one visible source of lag from animated resizing. The heavier SwiftUI settings tab rendering can be tracked separately.
> 
> 

Updated the PR description to scope this change to the resize animation.Thanks for the review.

PR #1188, opened 2026-05-24
refactor(stream): optimize chatStream iterator and error handling

MoonMao42:
# Summary
- Replaced the manual  .next()  call with a single  for try await  loop to prevent potential data loss or double-reading in the concurrent stream.

# Other
- Standardized error output: Stream errors are now formatted into standard JSON ( makeJSONErrorMessage ) before yielding to the client, preventing frontend parsing crashes caused by raw text errors.
- Cleaned up structure: Removed redundant  finish()  calls by utilizing  defer for cleaner and safer state management.


tisfeng (approved):
Good

PR #1196, opened 2026-06-04
fix(window): keep input window focused on multi-display and in-window clicks

superchun:
## Problem
On a multi-display setup, the input-translation (fixed) window kept losing
keyboard focus / getting dismissed while typing, interrupting input. It only
reproduced on certain spots and on the built-in display, not the external one.

## Root causes
1. **Click hit-test fails across displays.**
   `dismissWindowsIfMouseLocationOutsideFloatingWindow` used
   `window.frame.contains(NSEvent.mouseLocation)`. On multi-display / scaled
   setups the global mouse location and the window frame don't line up, so
   clicks *on* the window were judged "outside" and the unpinned window was
   dismissed mid-typing.
2. **Table steals first responder.**
   Clicking the result table body made the `NSTableView` first responder,
   pulling focus off the input text view.

## Fix
1. Use `NSWindow.windowNumber(at:)` (AppKit hit-testing) instead of manual
   `frame.contains`, so the click is matched to the real window under the
   cursor regardless of display arrangement or app-active state.
2. Set `tableView.refusesFirstResponder = YES` so clicks on the table body
   keep the input focused. Cell subviews (input field, selectable result text)
   still take first responder when clicked directly.

## Testing
Built locally (Xcode 26, macOS) and verified on a dual-display setup: clicking
anywhere on the input window no longer interrupts typing; clicking outside the
window still dismisses it as before.

### DIFF
--- a/Easydict/Swift/Utility/EventMonitor/Core/EventMonitor.swift
+++ b/Easydict/Swift/Utility/EventMonitor/Core/EventMonitor.swift
@@ -430,14 +430,23 @@ final class EventMonitor: NSObject {
     }
 
     private func dismissWindowsIfMouseLocationOutsideFloatingWindow(_ mouseLocation: CGPoint) {
-        if !checkIfMouseLocation(mouseLocation, in: EZWindowManager.shared().floatingWindow) {
+        guard let floatingWindow = EZWindowManager.shared().floatingWindow else {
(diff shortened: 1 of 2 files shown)


tisfeng (approved):
Looks good

